PR #975, opened 2014-12-20
Enable workspace cnf variables in project .bnd and .bndrun -runrequires (and others)

derheld42:
This change goes along with the companion bnd change. It enables
Workspace properties to be used in BndEditModel.

It also includes a R5LabelFormatter to show the variable on a runbnd
editor.

### DIFF
--- a/bndtools.core/src/bndtools/editor/BndEditor.java
+++ b/bndtools.core/src/bndtools/editor/BndEditor.java
@@ -72,6 +72,7 @@ import aQute.bnd.build.model.BndEditModel;
 import aQute.bnd.properties.BadLocationException;
 import bndtools.BndConstants;
 import bndtools.Plugin;
+import bndtools.central.Central;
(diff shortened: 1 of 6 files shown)


bjhargrave:
What is the corresponding bnd side commit?

pkriens:
I am confused, -runrequire can already handle macros? It is also already is a merge property. So can you indicate what you want to achieve?

derheld42:
I updated the title to be: Enable workspace cnf variables in project .bnd and .bndrun -runrequires (and others)

The original title was what I had when I initially implemented this before bnd mainline had the -runrequires variable support. It wasn't quite right.

As far as I can tell, the only piece still missing (from what my team currently uses) is utilizing variables from the workspace too.

Also -- The R5LabelFormatter change in this commit allows someone with a variable see that with the bnd/bndrun editors (rather than a parse error)

Issue #420, opened 2012-06-06
Access restriction due to required project


derheld42:
If you see this compile time error in eclipse and you're working with sub bundles:

<b>Access restriction: The constructor Foo() is not accessible due to restriction on required project bar</b>

This turns out to be an eclipse+BNDTools ism. The basic problem is that each eclipse project exports its "public interface", which for normal eclipse projects is all of the .class files it builds. This problem presents itself when using sub bundles, because you are creating multiple different public interfaces for a single eclipse project. There are cases where one of the definitions of the public interface is used, and it might be the wrong public interface needed for a bundle.

The fixes for this situation are:
1) http://www.digizol.org/2008/09/eclipse-access-restriction-on-library.html - Make the error into a warning and live with the warning.
2) Ensure you have the right set of -buildpath settings in the .bnd files you are using.
3) Restructure the sub bundles into bundles, so you have explicit eclipse projects for each bundle instead of one eclipse project with multiple sub bundles.
4) Figure out which project is causing the problematic interpretation of the public interface and add the sub bundle whose access has been restricted to the fore mentioned project.
5) Use a file.jar;version=file reference to the .bnd file instead of a bundle reference.
6) Add an eclipse jar file reference instead of a bundle reference in a .bnd file 

For my specific case of this problem, 2 worked. We had extraneous -buildpath settings that just needed to be cleaned up.
Also, I don't recommend workarounds 5 and 6.

It's not clear to me that this fundamental difference between Eclipse and BNDTools can be fixed. (However, I'm not an eclipse expert)

derheld42:
Bleh... I retested the test case up above and it no longer happens.

However, I have a case in my work environment which is still showing an access restriction issue. ugg. #984 fixes the issue I have in my work environment -- I verified that.

derheld42:
I retested the issue I had in my work enviroinment with pull request #999 and it resolves this issue. This can be closed out if #999 is accepted.

derheld42:
Fixed by #999

PR #984, opened 2014-12-23
Use property to control transitive dependencies in eclipse.

derheld42:
Someone could choose to workaround bndtools bug #420, where
transitive dependencies cause some issues inside eclipse.

This fix makes jar file references allow transitive dependencies.


derheld42:
#999 makes this obsolete.

PR #985, opened 2014-12-26
[launching] The run configuration clean overrode -runkeep

pkriens:
The  Run configuration specifies a clean option. This caused the setting of the -runkeep to be discarded. Will now only look at it when the -runkeep is off. Also removed the cleaning of the directory since this happens in the launcher.

Example of doing things in two places. This kind of stuff must only be done in bnd!


dhumeniuk:
Any plans on doing a bugfix release or other release with this fix in it soon?

bjhargrave:
It will be part of the 3.0 release this summer.

PR #991, opened 2015-01-04
Reducing number of RefreshFileJob threads

derheld42:
Improves performance when importing a lot of projects
and when refreshing a lot of projects.

### DIFF
--- a/bndtools.core/src/bndtools/central/RefreshFileJob.java
+++ b/bndtools.core/src/bndtools/central/RefreshFileJob.java
@@ -1,7 +1,10 @@
 package bndtools.central;
 
 import java.io.File;
+import java.util.Collections;
+import java.util.List;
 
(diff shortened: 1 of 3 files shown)


bndtools.core/src/bndtools/central/RefreshFileJob.java
```diff
@@ -14,42 +17,64 @@
...
-    private final IResource resource;
-    private final int depth;
+    private IResource resource;
+    private int depth;
 
     public RefreshFileJob(File file, boolean derived) throws Exception {
         super("Refreshing " + file);
         this.derived = derived;
+        this.files = new ArrayList<File>();
+        this.files.add(file);
```
bjhargrave: You should use Collections.singletonList(file) here instead of creating an array list and then adding the one file.

derheld42:
@bjhargrave - good catch. Updated.

PR #986, opened 2014-12-29
Jar Viewer: Expand/collapse directories on double click

seanbright:
Signed-off-by: Sean Bright <sean@malleable.com>

PR #992, opened 2015-01-07
Jar Viewer: Expand/collapse directories on double click

seanbright:


seanbright:
This is the same as this PR: https://github.com/bndtools/bndtools/pull/986

My testing environment was a bit wonky so I thought it wasn't working, but I was incorrect.

Issue #937, opened 2014-11-05
Project does not always build in Eclipse even though its build path is correct.


m4rr5:
Say you create three projects in a workspace:

 * pa - a project with build dependencies on pb and pc.a, and a class that uses a type A from pc.a
 * pb - a project with a build dependency on pc.b, and a class that uses a type B from pc.b
 * pc - a project with sub bundles, pc.a and pc.b

If you create the projects in this order, particularly with the build dependencies in pa in that order, Eclipse refuses to build pa because of the following error:

Access restriction: The type A is not accessible due to restriction on required project pc

If you now reverse the build order in pa to pc.a and then pb, it works.

In gradle, it works in all cases.

The problem seems to originate from the fact that Bndtools converts bundle dependencies into project dependencies with access restrictions. Furthermore, it seems that these project dependencies are transitive and that the first dependency "shadows" any following ones.

To reproduce, these are the exact projects and their files (I left out cnf, it is irrelevant to this setup, just generate it):

	pa/
		src/pa/UsesA.java
			package pa;
			import pc.a.A;
			public class UsesA {
				A a;
			}
		bnd.bnd
			-buildpath:  \
				pb;version=latest,\
				pc.a;version=latest
	pb/
		src/pb/UsesB.java
			package pb;
			import pc.b.B;
			public class UsesB {
				B b;
			}
		bnd.bnd
			-buildpath:  \
				pc.b;version=latest
	pc/
		src/pc/a/A.java
			package pc.a;
			public interface A {}
		src/pc/b/B.java
			package pc.b;
			public interface B {}
		a.bnd
			Export-Package: pc.a
		b.bnd
			Export-Package: pc.b
		bnd.bnd
			-sub: *.bnd

derheld42:
wow.. this is weird... I'm attempting to see if this is related to #999 and it's not, #999 deals with version=file references. However this is in the same area of code so I thought I'd take a swing at it.

I noticed this.. if you remove the pb;version=latest from the pa buildpath, it builds fine.
If you add it back, it doesn't work.

In BndContainerInitializer, it looks like the access rules are getting created correctly for the pc project. (i.e. you remove pb and it works fine). Somehow the pb reference -- which has a ** NOT ACCESSIBLE access rules, breaks the connection to pc. weird.

I started to fix it by not adding a project reference when there wasn't an export.. this kinda worked (although would have other negative sideeffects), and allowed things to build, until I added an export to pb, which then caused it to break again.

[...]

PR #999, opened 2015-01-12
Fix transition jar file references issue

derheld42:
Handles the case where:
Java Project A outputs a.jar containing package A
OSGI Project B has a version=file reference to a.jar and includes package A in export-packages.
OSGI Project C has buildpath reference to B and uses something in package A.

Without this change, Project C couldn't build due to
"not accessible due to restriction on required project" errors

### DIFF
--- a/bndtools.builder/src/org/bndtools/builder/classpath/BndContainerInitializer.java
+++ b/bndtools.builder/src/org/bndtools/builder/classpath/BndContainerInitializer.java
@@ -11,6 +11,7 @@ import java.util.LinkedHashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.StringTokenizer;
 import java.util.jar.Attributes.Name;
 import java.util.jar.Manifest;
@@ -47,6 +48,9 @@ import aQute.bnd.build.Container.TYPE;
 import aQute.bnd.build.Project;
 import aQute.bnd.build.Workspace;
 import aQute.bnd.header.Parameters;
+import aQute.bnd.osgi.Descriptors.PackageRef;
+import aQute.bnd.osgi.Jar;
+import aQute.bnd.osgi.Packages;
 import bndtools.central.Central;
 import bndtools.central.RefreshFileJob;
 
@@ -210,17 +214,31 @@ public class BndContainerInitializer extends ClasspathContainerInitializer imple
         }
 
         ArrayList<IClasspathEntry> result = new ArrayList<IClasspathEntry>(containers.size());
-        LinkedHashMap<Project,List<IAccessRule>> projectAccessRules = new LinkedHashMap<Project,List<IAccessRule>>();
+        LinkedHashMap<Project,List<IAccessRule>> projectAccessRulesExports = new LinkedHashMap<Project,List<IAccessRule>>();
+        LinkedHashMap<Project,List<IAccessRule>> projectAccessRulesPrivates = new LinkedHashMap<Project,List<IAccessRule>>();
         for (Container c : containers) {
             if (c.getType() == TYPE.PROJECT && c.getError() == null) {
-                calculateWorkspaceBundleAccessRules(projectAccessRules, c);
+                calculateWorkspaceBundleAccessRules(projectAccessRulesExports, projectAccessRulesPrivates, c, model);
+            }
+        }
+        //[cs] Add ** NO access to end of rule sets that have rules.
+        for (Entry<Project,List<IAccessRule>> accessrules : projectAccessRulesExports.entrySet()) {
+            if (accessrules.getValue() != null && accessrules.getValue().size() > 0) {
+                accessrules.getValue().add(JavaCore.newAccessRule(new Path("**"), IAccessRule.K_NON_ACCESSIBLE));
+            }
+        }
+        for (Entry<Project,List<IAccessRule>> accessrules : projectAccessRulesPrivates.entrySet()) {
+            if (accessrules.getValue() != null && accessrules.getValue().size() > 0) {
+                accessrules.getValue().add(JavaCore.newAccessRule(new Path("**"), IAccessRule.K_NON_ACCESSIBLE));
             }
         }
 
         List<File> filesToRefresh = new ArrayList<File>();
 
+        List<IAccessRule> exports = new LinkedList<IAccessRule>();
         for (Container c : containers) {
-            IClasspathEntry cpe;
+            IClasspathEntry cpe = null;
+            exports.clear();
 
             if (c.getError() == null) {
                 File file = c.getFile();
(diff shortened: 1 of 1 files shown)


derheld42:
This allows me to close out #420

bjhargrave:
How does this match the offline build behavior? Does this improve the fidelity?

derheld42:
I just tweaked the implementation. I forget to verify that non-exported packages shouldn't be accessible. I just tested that with the tweak.

I believe this brings bndtools closer to matching bnd. It definitely addresses (part of?) a problem with an eclipse project wrapping a .jar and providing portions of that jar as exported packages.

bndtools.builder/src/org/bndtools/builder/classpath/BndContainerInitializer.java
```diff
@@ -316,6 +324,27 @@ static boolean isProjectContainer(Container container) {
...
+                    if (tmp == null) {
+                        tmp = new LinkedList<IAccessRule>();
+                    }
+                    tmp.add(JavaCore.newAccessRule(new Path(pathStr), IAccessRule.K_ACCESSIBLE));
+                }
+            }
+        } catch (IOException e) {}
+        if (tmp != null && tmp.size() != 0) {
+            tmp.add(JavaCore.newAccessRule(new Path("**"), IAccessRule.K_NON_ACCESSIBLE));
+            return tmp.toArray(new IAccessRule[tmp.size()]);
```
bjhargrave: There is duplicate code here (add ** rule, convert to array). It would seem that it is better move that common code to the end of the method so all exit paths end here, no? That is, this method needs more refactoring for your change.

derheld42:
I tested another case:
Project A uses jar file B with packace C
Project A also overrides a class in package C
Project A exports package C
Project D has a buildpath reference to A.
Project D does sees the updated class from Project A.

The only thing that didn't work as expected was that I had to do a "rebuild" on project A to get the classpath to get recalculated so that Project D saw the right version of package C. This happens if package C was just created in Project A.

I think eclipse is providing the jar B package C instead of Project A package C. A rebuild cleared that up.

I'm sure there's a fix for to make it work right -- although I don't know if this is a likely case people will run into.

The basic problem is fixed with this pull request. There may be strange cases like this test right here that don't work "perfectly"

derheld42:
Refactored and rebased.

I ran into nastiness that slowed me down related to .gitattributes *.css and *.properties. I had to comment those 2 lines out. Otherwise everytime I did a git clean or git checkout, it would think I changed these two:
	modified:   bndtools.core/resources/unprocessed/intro/css/whatsnew.css
	modified:   bndtools.core/resources/unprocessed/intro/css/whatsnew.properties

Although now that I set core.autocrlf=false, it seems to have gone away. ugg.
I'm using git version 1.9.1 on ubuntu.

m4rr5:
Did you by any chance also validate if this PR fixes #937 as it sounds related.

bndtools.builder/src/org/bndtools/builder/classpath/BndContainerInitializer.java
```diff
@@ -303,16 +311,50 @@ static boolean isProjectContainer(Container container) {
...
+            // e.g.
+            // Project A generates A.jar with java package A
+            // Project B has a buildpath version=file reference to A.jar and exports package A
+            // Project C has a buildpath version=snapshot|latest reference to B and uses package A.
+            // The access rules calculated by this section permit Project C to
+            // use packages from Project A in eclipse.
+            Packages exportPkgs = model.getExports();
+            ZipFile zip = null;
+            try {
+                zip = new ZipFile(c.getFile());
```
bjhargrave: Instead of using ZipFile, I think it would be better to use aQute.bnd.osgi.Jar. You can then use the Jar object to interrogate the directories or package (which ever is the correct choice here).

bjhargrave:
> I ran into nastiness that slowed me down related to .gitattributes *.css and *.properties. I had to comment those 2 lines out. Otherwise everytime I did a git clean or git checkout, it would think I changed these two:
> modified: bndtools.core/resources/unprocessed/intro/css/whatsnew.css
> modified: bndtools.core/resources/unprocessed/intro/css/whatsnew.properties
> 
> Although now that I set core.autocrlf=false, it seems to have gone away. ugg.
> I'm using git version 1.9.1 on ubuntu.

I'll release a PR to fix this.

bjhargrave:
> I'll release a PR to fix this.

That issue is fixed by https://github.com/bndtools/bndtools/pull/1001.

derheld42:
I wonder if there's any good way to set up eclipse test projects for exercising weird stuff like this. hmm.

derheld42:
This last commit has a bit longer explanation for what this commit is now all about.

I now actually believe this will increase the fidelity between bndtools and bnd in terms of no-transitive dependencies. I think there might be a fringe case or 2 left, but this should resolve some of the other outstanding issues related to access issues and version=file references.

Outside of this issue, the NewBuilder doesn't seem to kick off a build when the classpath changes sometimes... and I don't know why. I traced through the NewBuilder code and it correctly updates the classpath, then returns a dependent project like it does elsewhere in the class. But a project rebuild is then required to get a project to build without errors. I have an example test case above in this issue if someone is interested.

bndtools.builder/src/org/bndtools/builder/classpath/BndContainerInitializer.java
```diff
@@ -255,26 +273,59 @@ public static void setClasspathEntries(IJavaProject javaProject, IClasspathEntry
...
         }
 
+        //[cs] set project variable: "accessrulesdebug: true" to enable some extra eclipse output for debugging.
+        if (Boolean.parseBoolean(model.getProperty("accessrulesdebug", "false"))) {
+            System.out.println("IClasspathEntrys: project = " + model.getName());
+            for (IClasspathEntry f : result) {
+                System.out.println("--- " + f);
+            }
+        }
+
```
bjhargrave: I assume the above code should not be part of the final PR?

bndtools.builder/src/org/bndtools/builder/classpath/BndContainerInitializer.java
```diff
@@ -210,17 +214,31 @@ public static void setClasspathEntries(IJavaProject javaProject, IClasspathEntry
...
-        LinkedHashMap<Project,List<IAccessRule>> projectAccessRules = new LinkedHashMap<Project,List<IAccessRule>>();
+        LinkedHashMap<Project,List<IAccessRule>> projectAccessRulesExports = new LinkedHashMap<Project,List<IAccessRule>>();
+        LinkedHashMap<Project,List<IAccessRule>> projectAccessRulesPrivates = new LinkedHashMap<Project,List<IAccessRule>>();
         for (Container c : containers) {
             if (c.getType() == TYPE.PROJECT && c.getError() == null) {
-                calculateWorkspaceBundleAccessRules(projectAccessRules, c);
+                calculateWorkspaceBundleAccessRules(projectAccessRulesExports, projectAccessRulesPrivates, c, model);
+            }
+        }
+        //[cs] Add ** NO access to end of rule sets that have rules.
```
bjhargrave: Why is there "[cs]" in front of all the comments? git blame can tell us you write these comments.

bndtools.builder/src/org/bndtools/builder/classpath/BndContainerInitializer.java
```diff
@@ -342,12 +416,26 @@ static void calculateWorkspaceBundleAccessRules(Map<Project,List<IAccessRule>> p
...
+                String exportDir = exportPkg.replace('.', '/');
+                String pathStr = exportDir + "/*";
+                privates.add(JavaCore.newAccessRule(new Path(pathStr), IAccessRule.K_ACCESSIBLE));
+                for (PackageRef containedPackage : contained.keySet()) {
+                    String dir = containedPackage.getBinary();
+                    if (dir.equals(exportDir)) {
+                        exports.add(JavaCore.newAccessRule(new Path(pathStr), IAccessRule.K_ACCESSIBLE));
+                    }
+                }
+                //TODO -- fix bnd bug.
```
bjhargrave: What is the bug here?

bndtools.builder/src/org/bndtools/builder/classpath/BndContainerInitializer.java
derheld42: I thought it was useful for debugging purposes. I keep adding code like that to aid debugging. Might as well add it to the baseline.

bndtools.builder/src/org/bndtools/builder/classpath/BndContainerInitializer.java
derheld42: Habit of mine. It's a quick way of knowing who added the comment. I find it useful. If the bndtools coding standard says don't do that, I can remove it.

bndtools.builder/src/org/bndtools/builder/classpath/BndContainerInitializer.java
derheld42: I think so. Looks like copy and paste bug. The containsbinaryname method seems wrong. I didn't write anything up yet.

bndtools.builder/src/org/bndtools/builder/classpath/BndContainerInitializer.java
bjhargrave: My opinion would be to leave it out.

bndtools.builder/src/org/bndtools/builder/classpath/BndContainerInitializer.java
derheld42: @pkriens is there a copy 'n paste bug in aQute.bnd.osgi.Packages.containsBinaryName ? It has the same implementation as containsFQN()

bndtools.builder/src/org/bndtools/builder/classpath/BndContainerInitializer.java
derheld42: I switched it to a logger in the updated commit.

derheld42:
Any concerns about merging this in? It should see some bake time in the baseline before another release.

I know I have the power to merge it in myself, but I'm not quite sure when a review must/should happen... What's our protocol?

I suppose a "bad" change could always get reverted... but is that how we're working?

pkriens:
It is rather simple :-) 

If you make a simple fix, merge it yourself. If it is in any way imaginable contentious (workspace layout for example :-), you feel unsure, or it is large then ping the group and ask for review. As you said, it is easy to undo.

In this case, go for it. Looks like you did some serious work in this black art part of bndtools ...

bndtools.builder/src/org/bndtools/builder/classpath/BndContainerInitializer.java
bjhargrave: Why don't you open a bnd issue for this? It can be tracked separately.

derheld42: Done. bndtools/bnd#782 written.

derheld42:
doh... I missed something. I'm either going to fix it or revert it.

pkriens:
:-)

PR #1002, opened 2015-01-17
Revert "Fix transitive jar file references issues"

derheld42:
This reverts commit dfa63eb06defb1a131c9efd35c49efc266f21aca.
Otherwise known as #999 

Turns out I didn't test enough until AFTER I pushed it.

I need to think about how to solve eclipse's squirreliness.

Issue to solve:
- bnd has projects that pulls packages from other projects and exports them. And there are other projects that have references to the project that did the pulling. My approach didn't address this. (but solved other issues!) <grumble, grumble>


derheld42:
For the record... The case that seems to be blowing up in bnd... I had a test case just like it. I even adjusted my test case to "match".. and it works...

So, there's some other subtlety going on. joy.

PR #1003, opened 2015-01-19
Created a set of new access rules for eclipse

derheld42:
These are not turned on by default, since I wanted to slow roll the incorporation
of something so fundamental to bndtools.

To enable, add a project or global variable:
eclipse.newaccessrules: true

You can also debug access rules with
eclipse.debug: true

The new rules add library classpath entries instead of project classpath entries
except for version=project references, which still add project references. All
transitive dependencies are turned off with the new rules too. This should be
closer to the bnd command line build handling of jars on the classpath during
a build.

These new access rules should address several transitive dependency issues that
I've encountered and I've seen written up as issues.

### DIFF
--- a/bndtools.builder/src/org/bndtools/builder/classpath/BndContainerInitializer.java
+++ b/bndtools.builder/src/org/bndtools/builder/classpath/BndContainerInitializer.java
@@ -11,6 +11,7 @@ import java.util.LinkedHashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.StringTokenizer;
 import java.util.jar.Attributes.Name;
 import java.util.jar.Manifest;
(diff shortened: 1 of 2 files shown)


derheld42:
After thinking about BJ's comment -- https://groups.google.com/forum/#!topic/bndtools-dev/ogng_UleV5w

I came up with something that seems to work in bnd, bndtools, and 2 other test cases I've been chasing. It solves #420 and #937. 

Rather than changing something so fundamental in bndtools outright, I thought I'd add a variable that someone can turn on to enable the new rules. This way it can be slow rolled. If we determine it's right, we can change the default of that variable down the road.

Thoughts? Any issues with merging this in?

bjhargrave:
Looks good to me. Lets merge this in and then we should start using it with `eclipse.newaccessrules: true`. We should consider setting that in the bnd and bndtools cnf/build.bnd to get internal experience with this.

bjhargrave:
Looking good! I tried this on a cleaned bnd (and also bndtools) repo imported into a fresh eclipse workspace. This meant bndtools had to compile the repo upon import to create the generated jars referenced by other projects so they could compile. It was successful.

Also, I used the new access rules to build a cleaned OSGi workspace (> 160 projects) and that worked also. So this is looking very good to me! Congrats.

You should also open a issue to remind us to change the code to only the new access rules for the 3.0 release.

derheld42:
wrote #1005 to remind us down the road.

PR #1008, opened 2015-01-22
Allow mulitple git repos in one eclipse workspace

rimmeraj:
This change along with https://github.com/bndtools/bnd/pull/704
allows multiple git repos in the eclipse workspace. It does this by passing in the file path of each project instead of just the name. That way bnd does not have to assume that the project is relative to the cnf folder. The refresh task was changed to ask bnd if the file path was in one of the projects since all the projects are no longer in the same root workspace.

LaunchUtils.java, ReleaseAction.java were changed to use the getProject helper for consistency.

Signed-off-by: Dave Smith <dave.smith@candata.com>


bjhargrave:
Closing per https://github.com/bndtools/bnd/pull/704#issuecomment-71533322

PR #1062, opened 2015-03-06
Additions to persistent dcorators

pkriens:
To be discussed with BJ

I did almost the identical work BJ did last night but I handle extra source folders and substitutable imports. Let's discuss.


bjhargrave:
Replaced by #1063.

PR #1132, opened 2015-07-08
Add vertical scrollbar to resolution-error-textarea

lostiniceland:
I added the vertical scrollbar to the textarea. 

A demo-workspace was added as well with an single jar that will fail the resolution. Just hit resove and resize the window to be small enough


bjhargrave:
You need to redo your commit to (a) add a Signed-off-by to the commit messages and (b) remove the workspace-demo folder from your commit.

PR #1150, opened 2015-08-22
Updated bndtools.shared.bndrun to work with Eclipse 4.3.2

gamerson:


bjhargrave:
I updated the bintray repo with the files you requested from the delta pack. The index was also updated to include these new files.

bjhargrave:
@gamerson Please merge when you have confirmed the bintray update is correct. Thanks.

gamerson:
@bjhargrave  looks good!  bndrun launch found the jars from bintray, thanks for the fast turnaround.

PR #1157, opened 2015-08-28
Update DS annotation option text to reflect 1.3

seanbright:
Not in love with this, but with 1.3 generally available it is a bit more
clear about capabilities.

### DIFF
--- a/bndtools.core/src/bndtools/editor/contents/GeneralInfoPart.java
+++ b/bndtools.core/src/bndtools/editor/contents/GeneralInfoPart.java
@@ -88,7 +88,7 @@ public class GeneralInfoPart extends SectionPart implements PropertyChangeListen
     };
 
     private static enum ComponentChoice {
-        None("None/Undefined"), Bnd("Bnd Annotations"), DS("DS 1.2 Annotations");
+        None("None/Undefined"), Bnd("Bnd Annotations"), DS("OSGi DS Annotations");
 
         private final String label;
 


bjhargrave:
In bnd 3.0, we changed the default for `-dsannotations` to be `*`. So you do not need to set `-dsannotations: *` in your bnd file anymore to get OSGi DS annotations processing. So I am not sure what the value of the UI is anymore.

Also, if we leave this UI, I don't care for `1.2+`. It would be better as simply `OSGi DS Annotations`.

seanbright:
Does `-dsannotations: *` process both the bnd and the DS annotations? If not, we need the UI element to allow for people that don't want to update to the standard annotations.

bjhargrave:
> Does -dsannotations: * process both the bnd and the DS annotations? 

No, just the OSGi DS annotations.

> If not, we need the UI element to allow for people that don't want to update to the standard annotations.

We want to nudge people off the bnd annotations. In the 3.0 release note, we announce the intention to deprecate the bnd annotations in a future release. So I am not sure how this will affect the UI.

seanbright:
It's a minor change either way, so feel free to the close this PR if this isn't the direction you want to go.

seanbright:
This a dead-end?

seanbright:
Sorry to be a pest :grimacing:

bjhargrave:
Squeaky wheel...

seanbright:
Message received.

PR #1158, opened 2015-08-28
Don't run the resolver if there is no point

seanbright:
This disables the 'Resolve' button if we have an empty list of Run Requiments and also checks on save (if auto resolve is enabled) before running the resolver.


seanbright:
Will think on this

PR #1182, opened 2015-09-18
Use OSGi annotations in New OSGi Declarative Services Component wizard

seanbright:


bjhargrave:
Isn't more needed? Like changing the project's buildpath to remove bnd annotations jar and putting in the osgi R6 cmpn jar?

seanbright:
You're right. I will revisit this when I've done more testing. Sorry for the noise.

PR #1222, opened 2015-11-20
1160: use Eclipse-Job-API for repository actions

lostiniceland:
To give the user some visual feedback for long running
repository-actions (e.g. "Download All") the Runnable is wrapped with
the Eclipse-Job-API.

This will only give the user feedback that there is a job running and when it is done. 
In order to add more feedback with a progressbar the bnd-side must be enhanced with some other interface than Runnable (maybe some simple callback-interface which reveives an int for the current status).

### DIFF
--- a/bndtools.core/src/bndtools/views/repository/RepositoriesView.java
+++ b/bndtools.core/src/bndtools/views/repository/RepositoriesView.java
@@ -71,7 +71,6 @@ import org.eclipse.swt.layout.GridData;
 import org.eclipse.swt.layout.GridLayout;
 import org.eclipse.swt.widgets.Composite;
 import org.eclipse.swt.widgets.Control;
-import org.eclipse.swt.widgets.Display;
 import org.eclipse.swt.widgets.Menu;
 import org.eclipse.swt.widgets.Tree;
 import org.eclipse.ui.ISharedImages;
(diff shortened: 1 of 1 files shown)


bjhargrave:
Please review @njbartlett.

bndtools.core/src/bndtools/views/repository/RepositoriesView.java
```diff
@@ -613,17 +618,42 @@ public void menuAboutToShow(IMenuManager manager) {
...
+                                            Job backgroundJob = new Job("Repository Action '" + getText() + "'") {
+
+                                                @Override
+                                                protected IStatus run(IProgressMonitor monitor) {
+                                                    try {
+                                                        e.getValue().run();
+                                                        if (rp != null && rp instanceof Refreshable)
+                                                            Central.refreshPlugin((Refreshable) rp);
+                                                    } catch (final Exception e) {
+                                                        Display.getDefault().asyncExec(new Runnable() {
```
bjhargrave: Should this really be a dialog? Or just put in the error log? I doubt there is much the user can do here except dismiss the dialog.

lostiniceland: Right, the error-log is the better place to put this.

bjhargrave:
Can you please squash into a single commit without a merge commit? If that exceeds your git fu, let me know and I can do it.

PR #1223, opened 2015-11-22
796: Run Requirements: don't allow duplicates

lostiniceland:
Requirements are checked with "equals" provided by BND. 
Ignored Requirements are displayed as information-message.

### DIFF
--- a/bndtools.core/src/bndtools/editor/project/RunRequirementsPart.java
+++ b/bndtools.core/src/bndtools/editor/project/RunRequirementsPart.java
@@ -454,23 +454,18 @@ public class RunRequirementsPart extends BndEditorPart implements PropertyChange
     }
 
     /**
-     * Update the requirements already available with new ones. Requirements are compared with the ones already in the
-     * viewer.
+     * Update the requirements already available with new ones. Already existing requirements will be removed from the
+     * given set.
      *
      * @param adding
      *            Set with {@link Requirement}s to add
      */
     private void updateViewerWithNewRequirements(Set<Requirement> adding) {
-        final List<Requirement> finalResult = new ArrayList<>(adding.size());
-        // check for duplicates
-        for (Requirement req : adding) {
-            if (!requires.contains(req)) {
-                finalResult.add(req);
-            }
-        }
-        if (!finalResult.isEmpty()) {
-            requires.addAll(finalResult);
-            viewer.add(finalResult.toArray(new Object[finalResult.size()]));
+        // remove duplicates
+        adding.removeAll(requires);
+        if (!adding.isEmpty()) {
+            requires.addAll(adding);
+            viewer.add(adding.toArray(new Object[adding.size()]));
             markDirty();
         }
     }
--- a/bndtools.core/src/bndtools/editor/project/RunRequirementsPart.java
+++ b/bndtools.core/src/bndtools/editor/project/RunRequirementsPart.java
@@ -5,9 +5,11 @@ import java.beans.PropertyChangeListener;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.EnumSet;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Set;
 
 import org.bndtools.api.ILogger;
 import org.bndtools.api.Logger;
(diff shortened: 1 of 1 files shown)


bjhargrave:
Please review @njbartlett.

bndtools.core/src/bndtools/editor/project/RunRequirementsPart.java
```diff
@@ -269,11 +271,7 @@ private void doAddBundle() {
                     Requirement req = new CapReqBuilder(IdentityNamespace.IDENTITY_NAMESPACE).addDirective(Namespace.REQUIREMENT_FILTER_DIRECTIVE, filter.toString()).buildSyntheticRequirement();
                     adding.add(req);
                 }
-                if (!adding.isEmpty()) {
```
bjhargrave: Here you remove the empty check before calling the new method. Below you keep the empty check. I would assume these should be consistent.

bndtools.core/src/bndtools/editor/project/RunRequirementsPart.java
```diff
@@ -474,9 +512,7 @@ protected boolean performSelectionDrop(ISelection data, Object target, int locat
             }
 
             if (!adding.isEmpty()) {
```
bjhargrave: Here you leave the empty check before calling the new method. Above you remove the empty check. I would assume these should be consistent.

bndtools.core/src/bndtools/editor/project/RunRequirementsPart.java
```diff
@@ -257,8 +259,8 @@ private void doAddBundle() {
             WizardDialog dialog = new WizardDialog(getSection().getShell(), wizard);
 
             if (Window.OK == dialog.open()) {
-                List<VersionedClause> result = wizard.getSelectedBundles();
-                List<Requirement> adding = new ArrayList<Requirement>(result.size());
+                final List<VersionedClause> result = wizard.getSelectedBundles();
+                final List<Requirement> adding = new ArrayList<Requirement>(result.size());
```
bjhargrave: Couldn't this be much simpler if `adding` was a Set? Then duplicate adds would just be ignored.

bndtools.core/src/bndtools/editor/project/RunRequirementsPart.java
lostiniceland: This was on purpose. The intention was to remove the empty-check, but it was still necessary in the DropAdapter due to the boolean-return.

Going to add the check above for consistency.

bjhargrave: Actually, I think the best solution is to simply replace the List with a Set. We don't need a dialog popup.

bndtools.core/src/bndtools/editor/project/RunRequirementsPart.java
lostiniceland: This depends on what you would like to display to the user:

If we dont want to display the duplicates to the user in the dialog, I can make this a Set. I decided to go for separate lists in order to give some feedback. Even though the dialog is not pretty and might anoy somebody I thought it to be better than a user who is wondering why nothing happened (editor marked dirty) after he added bundles.

Best solution might be to check if a bundle can be added during the selection but this has to be done in multiple places, so I went for the approach with the smallest impact.

bjhargrave: Just do `adding.removeAll(requires)` to remove any existing requirements before the empty check. Then the editor will only be marked dirty if there are new requirements being added.

bndtools.core/src/bndtools/editor/project/RunRequirementsPart.java
```diff
@@ -270,9 +272,7 @@ private void doAddBundle() {
                     adding.add(req);
                 }
                 if (!adding.isEmpty()) {
```
bjhargrave: You do the adding check in the new method, so remove it here.

bndtools.core/src/bndtools/editor/project/RunRequirementsPart.java
```diff
@@ -453,6 +453,23 @@ public void propertyChange(PropertyChangeEvent evt) {
...
     }
 
+    /**
+     * Update the requirements already available with new ones. Already existing requirements will be removed from the
+     * given set.
+     *
+     * @param adding
+     *            Set with {@link Requirement}s to add
+     */
+    private void updateViewerWithNewRequirements(Set<Requirement> adding) {
```
bjhargrave: Have the method return boolean: !adding.isEmpty().

bndtools.core/src/bndtools/editor/project/RunRequirementsPart.java
```diff
@@ -474,9 +491,7 @@ protected boolean performSelectionDrop(ISelection data, Object target, int locat
             }
 
             if (!adding.isEmpty()) {
```
bjhargrave: Remove the isEmpty check and just return the result of the new method.

bjhargrave:
Also please squash into a single commit. Thanks!

bjhargrave:
I moved this work to https://github.com/bndtools/bndtools/pull/1253

PR #1232, opened 2015-12-04
New Template System

njbartlett:
@pkriens Please check basic functionality of enRoute templates as discussed.


njbartlett:
I have run through all the magic enRoute project names -- `api`, `application`, `example`, `provider`, `test` and `webresource`. All seem to generate something sensible and there are no build error markers. I haven't tried running anything from these yet, however.

Unless I hear otherwise by Monday, I will merge this PR. Of course we can still fix bugs and clean up code during the RC phase.

pkriens:
Let's merge this and see how it looks like. I think it looks ok but it is hard to oversee the effects of this large PR. As far as I could see you still support the old extension points so that would void my concerns. Thanks.

njbartlett:
Yeah I wasn't really asking for a review of the code, just check whether the output from the templates is correct.

Anyway, merging.

PR #1258, opened 2015-12-17
Jar Viewer: Sort the tree

seanbright:
Directories first by name, and then files by name.


seanbright:
Can obviously wait for 3.2

bjhargrave:
I am not sure it is wise to sort the view of the zip contents. The order in the zip matters (which is why a LinkedHashMap was used). For example Jar spec says META-INF must be first. If we offer  to sort the view, then perhaps there should be a UI element on the viewer which allows me to request sorted or actual order.

PR #1264, opened 2015-12-24
Add DnD support for advanced search results.

### LINKED ISSUE #1230: Can't drag-and-drop from results of advanced search
When using the advanced search in the repositories view, e.g. to search for a package, I cannot drag and drop from the search results to the build path dependencies section.

seanbright:
I'm not convinced that this is the most elegant way of implementing
this, so feedback is certainly appreciated.

Fixes #1230

Signed off by: Sean Bright <sean.bright@gmail.com>


seanbright:
I've assigned to Neil because the advanced search was his work and I want to make sure I'm not mucking things up.

Issue #1274, opened 2016-01-27
Ambiguous warnings from ProjectPathsValidator


jawi:
When updating our projects to the latest greatest (3.1.1) Bndtools, we get a set of warnings that are rather ambiguous, for example:

    Eclipse: The .classpath contains an unknown container: org.eclipse.jdt.junit.JUNIT_CONTAINER/4

The message looks like as if Eclipse itself is complaining about this instead of Bndtools. In addition, the message suggests that the container itself is no longer recognised. Though I've doubts whether this is actually a valid warning, I propose that the message (and the other messages generated by the validator) should state something like:

    Bndtools: the classpath contains an unknown container: <NAME>. This could make your build not portable.

bjhargrave:
Sounds good. Can you make a PR?

jawi:
Will do.

jawi:
PR created in #1275.

PR #1275, opened 2016-01-28
Issue #1274 - less ambiguous validation messages:

jawi:
- make it clear that the validation message comes from Bndtools and not
  from Eclipse;
- make it clear that an unknown container might impact your build's
  portability;
- small typo fixed.

Signed-off-by: Jan Willem Janssen <janwillem.janssen@luminis.eu>


bjhargrave:
I deleted this since it was against next and not master. I did not notice until after I hit merge. But I have cherry-picked this onto master with #1276.

PR #1296, opened 2016-02-25
with-Eclipse logo

paulvi:
and this is developed with Eclipse


bjhargrave:
I am not sure why we would add a link to your website to our readme. I don't see the purpose or value in this. If we wanted to promote an associate with the Eclipse Foundation, we would to their website and use their logo. You logo is out-of-date with respect to the current Eclipse logo design.

Closing.

PR #1346, opened 2016-04-19
ignore property changes when committing the model fixes #1297

### LINKED ISSUE #1297: Launcher/JVM Arguments vanish
I'm trying to edit the .bndrun file for adding Launcher/JVM Arguments. Adding them on the Run tab makes them disappear once I save. Adding the properties on the Source tab make them appear on the Run tab correctly. Seems like they get lost on save somehow.

I'm running bndtools 3.1.1 on Neon M5

gamerson:


bndtools.core/src/bndtools/editor/common/BndEditorPart.java
```diff
@@ -17,6 +17,7 @@
 
     protected BndEditModel model;
 
+    private boolean committing = false;
```
bjhargrave: Does this need to be `volatile` or `AtomicBoolean`? Not sure if `commit` and `propertyChange` are called on the same thread.

gamerson: I don't think its safe to assume that it would only be modified from UI Thread, I can switch it to AtomicBoolean.

bjhargrave:
Looks good. Can you squash the commits?

PR #1347, opened 2016-04-19
If user double-clicked a RepositoryEntry that is not downloaded ask them if they wish to download, if so, grab the file and open it

gamerson:


bndtools.core/src/bndtools/views/repository/RepositoriesView.java
```diff
@@ -303,6 +303,16 @@ public void doubleClick(DoubleClickEvent event) {
                     Object element = selection.getFirstElement();
                     if (element instanceof IAdaptable) {
                         URI uri = (URI) ((IAdaptable) element).getAdapter(URI.class);
+                        if (uri == null && element instanceof RepositoryEntry) {
+                            boolean download = MessageDialog.openQuestion(getSite().getShell(), "Repositories", "This repository entry is unable to be opened because it has not been downloaded. " + "Download and open it now?");
+                            if (download) {
```
bjhargrave: Should the download occur here on the UI thread? Perhaps it should be pushed into a background job?

gamerson: Sure, we can do background job.

bjhargrave:
LGTM

PR #1349, opened 2016-04-21
Check for null Project to avoid NPE

gamerson:


gamerson:
This small null check helps the bndEditor be used to edit a bnd.bnd file to set "Export-Package:" from a non-workspace project.

bjhargrave:
LGTM

PR #1368, opened 2016-05-13
Add real bundle versions into feature.xml

gamerson:


gamerson:
For normal Eclipse installations having actual version numbers versus 0.0.0 is not a big difference.  Eclipse can handle 0.0.0.  However, for people who are building solutions with Eclipse Tycho and they want to include bndtools in the final repository, we can't just simply point to the bndtools.main.feature because Tycho must not be following the 0.0.0 to get the actual plugins. So this PR just enumerates the versions in the feature.xml, which makes bndtools's feature.xml a little bit more standard.

build/build.gradle
```diff
@@ -112,29 +112,19 @@ void generateP2Repository(String title, String antTarget, String metadataReposit
...
-  dependsOn p2FeatureTree
-  destinationDir = file("${buildDir}/features")
-  archiveName    = 'bndtools.feature.jar'
-  from             "${buildDir}/features/bndtools"
-  include          'feature.xml'
+  def tokens = ['master-version' : masterVersion, 'base-version' : bnd('base-version')]
+  def plugins = fileTree("${buildDir}/plugins").getFiles()
+  plugins.each {
+    Jar jar = new Jar(it)
+    tokens.put(jar.getBsn() + '-version', jar.getVersion())
```
bjhargrave: You need to close these jars. When gradle runs in daemon mode, it can hold them open for a long time and Windows hates open files...

```groovy
new Jar(it).withCloseable { jar ->
  tokens.put(jar.getBsn() + '-version', jar.getVersion())
}
```

gamerson: oh nasty, ok, wil repush

PR #1392, opened 2016-06-06
[progress] display bnd downloading tasks in Eclipse progress view

gamerson:
This requires [some changes](https://github.com/bndtools/bnd/pull/1492) to bnd in order to properly display in Eclipse Progress View.

Here is a screenshot of this in action
bndtools-progress

PR #1393, opened 2016-06-07
Use progress view and reduce UI thread usage

gamerson:
This requires some changes to bnd in order to work properly.  https://github.com/bndtools/bnd/pull/1492

I just wanted to get some early feedback and also a UI review since this changes the UX a lot.  Attaching a screenshot
workspace-is-loading

### DIFF
--- a/bndtools.builder/src/org/bndtools/builder/classpath/BndContainerInitializer.java
+++ b/bndtools.builder/src/org/bndtools/builder/classpath/BndContainerInitializer.java
@@ -38,9 +38,12 @@ import org.eclipse.jdt.core.IJavaProject;
 import org.eclipse.jdt.core.JavaCore;
 import org.eclipse.jdt.core.JavaModelException;
 import org.eclipse.jdt.internal.core.JavaModelManager;
+import org.osgi.util.function.Function;
+
 import aQute.bnd.build.CircularDependencyException;
 import aQute.bnd.build.Container;
 import aQute.bnd.build.Project;
+import aQute.bnd.build.Workspace;
 import aQute.bnd.header.Parameters;
 import aQute.bnd.osgi.Constants;
 import aQute.bnd.osgi.Descriptors.PackageRef;
@@ -61,15 +64,25 @@ public class BndContainerInitializer extends ClasspathContainerInitializer imple
 
     public BndContainerInitializer() {
         super();
-        Central.getInstance().addModelListener(this);
     }
 
     @Override
-    public void initialize(IPath containerPath, IJavaProject javaProject) throws CoreException {
-        IProject project = javaProject.getProject();
+    public void initialize(IPath containerPath, final IJavaProject javaProject) throws CoreException {
+        Central.onWorkspaceInit(new Function<Workspace,Void>() {
+            @Override
+            public Void apply(Workspace a) {
+                try {
+                    Central.getInstance().addModelListener(BndContainerInitializer.this);
+                    IProject project = javaProject.getProject();
 
-        Updater updater = new Updater(project, javaProject);
-        updater.updateClasspathContainer(true);
+                    Updater updater = new Updater(project, javaProject);
+                    updater.updateClasspathContainer(true);
+                } catch (CoreException e) {
+                    logger.logError("Error initializing classpath container", e);
+                }
+                return null;
+            }
+        });
     }
 
     @Override
--- a/bndtools.builder/src/org/bndtools/builder/classpath/BndContainerInitializer.java
+++ b/bndtools.builder/src/org/bndtools/builder/classpath/BndContainerInitializer.java
@@ -64,6 +64,13 @@ public class BndContainerInitializer extends ClasspathContainerInitializer imple
 
     public BndContainerInitializer() {
         super();
+        Central.onWorkspaceInit(new Function<Workspace,Void>() {
+            @Override
+            public Void apply(Workspace t) {
+                Central.getInstance().addModelListener(BndContainerInitializer.this);
+                return null;
+            }
+        });
     }
 
     @Override
@@ -72,7 +79,6 @@ public class BndContainerInitializer extends ClasspathContainerInitializer imple
             @Override
             public Void apply(Workspace a) {
                 try {
-                    Central.getInstance().addModelListener(BndContainerInitializer.this);
                     IProject project = javaProject.getProject();
 
                     Updater updater = new Updater(project, javaProject);
--- a/bndtools.core/src/bndtools/central/Central.java
+++ b/bndtools.core/src/bndtools/central/Central.java
@@ -18,7 +18,6 @@ import org.bndtools.api.ILogger;
 import org.bndtools.api.IStartupParticipant;
 import org.bndtools.api.Logger;
 import org.bndtools.api.ModelListener;
-import org.bndtools.utils.Function;
 import org.eclipse.core.resources.IFile;
 import org.eclipse.core.resources.IProject;
 import org.eclipse.core.resources.IResource;
@@ -42,6 +41,7 @@ import org.eclipse.jface.viewers.TreeViewer;
 import org.osgi.framework.Bundle;
 import org.osgi.framework.BundleContext;
 import org.osgi.framework.FrameworkUtil;
+import org.osgi.util.function.Function;
 
 import aQute.bnd.build.Project;
 import aQute.bnd.build.Workspace;
@@ -220,15 +220,17 @@ public class Central implements IStartupParticipant {
             // Monitor changes in cnf so we can refresh the workspace
             addCnfChangeListener(newWorkspace);
 
-            // The workspace has been initialized fully, set the field now
-            workspace = newWorkspace;
-
-            workspaceRepositoryChangeDetector = new WorkspaceRepositoryChangeDetector(workspace);
+            workspaceRepositoryChangeDetector = new WorkspaceRepositoryChangeDetector(newWorkspace);
 
             // Call the queued workspace init callbacks
-            while (!workspaceInitCallbackQueue.isEmpty()) {
-                Function<Workspace,Void> callback = workspaceInitCallbackQueue.remove(0);
-                callback.run(workspace);
+            synchronized (workspaceInitCallbackQueue) {
+                // The workspace has been initialized fully, set the field now
+                workspace = newWorkspace;
+
+                while (!workspaceInitCallbackQueue.isEmpty()) {
+                    Function<Workspace,Void> callback = workspaceInitCallbackQueue.remove(0);
+                    callback.apply(workspace);
+                }
             }
 
             return workspace;
@@ -240,11 +242,13 @@ public class Central implements IStartupParticipant {
         }
     }
 
-    public synchronized static void onWorkspaceInit(Function<Workspace,Void> callback) {
-        if (workspace != null)
-            callback.run(workspace);
-        else
-            workspaceInitCallbackQueue.add(callback);
+    public static void onWorkspaceInit(Function<Workspace,Void> callback) {
+        synchronized (workspaceInitCallbackQueue) {
+            if (workspace != null)
+                callback.apply(workspace);
+            else
+                workspaceInitCallbackQueue.add(callback);
+        }
     }
 
     private static File getWorkspaceDirectory() throws CoreException {
--- a/bndtools.core/src/bndtools/central/Central.java
+++ b/bndtools.core/src/bndtools/central/Central.java
@@ -3,10 +3,10 @@ package bndtools.central;
 import java.io.File;
 import java.util.ArrayList;
 import java.util.HashMap;
-import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.Callable;
+import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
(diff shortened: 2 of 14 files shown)


bjhargrave:
Does this replace #1392?

bndtools.builder/src/org/bndtools/builder/BndtoolsBuilder.java
```diff
@@ -117,6 +119,7 @@
             if (model == null) {
                 try {
                     model = Central.getProject(myProject.getLocation().toFile());
+                    model.addBasicPlugin(new JobProgress());
```
bjhargrave: The plugins are a set and you keep adding a new one every time a project is processed by the builder. This seems like a memory leak.

bndtools.core/src/bndtools/types/Function.java
```diff
@@ -1,4 +1,4 @@
-package org.bndtools.utils;
+package bndtools.types;
```
bjhargrave: Perhaps this is an old type, but why don't we use org.osgi.util.function.Function? bnd now uses Promises so this type is around.

gamerson: Thanks I'll switch to org.osgi.util.function.Function

bndtools.builder/src/org/bndtools/builder/BndtoolsBuilder.java
gamerson: Yes, i just checked and the model that is returned already has a JobProgress plugin (probably inherited from workspace)

gamerson:
I merged previous PR and review commits, squashed and repushed here.

bndtools.builder/src/org/bndtools/builder/classpath/BndContainerInitializer.java
```diff
@@ -61,15 +64,25 @@
 
     public BndContainerInitializer() {
         super();
-        Central.getInstance().addModelListener(this);
```
bjhargrave: Why remove this call? The class still implements ModelListener.

bndtools.builder/src/org/bndtools/builder/classpath/BndContainerInitializer.java
```diff
@@ -61,15 +64,25 @@
...
 
     @Override
-    public void initialize(IPath containerPath, IJavaProject javaProject) throws CoreException {
-        IProject project = javaProject.getProject();
+    public void initialize(IPath containerPath, final IJavaProject javaProject) throws CoreException {
+        Central.onWorkspaceInit(new Function<Workspace,Void>() {
+            @Override
+            public Void apply(Workspace a) {
+                try {
+                    Central.getInstance().addModelListener(BndContainerInitializer.this);
```
bjhargrave: I am not sure if this solves the problem I was raising. Eclipse only creates a single instance of BndContainerInitializer. Not one per projects. So the constructor is only called once and the model listener is only added once. This change calls addModelListener multiple times. But since the method wont add if already added, it is not harmful. However, I think the cleaner solution is to make an onWorkspaceInit call in the constructor to add the model listener rather than piggy backing on the initialize call which is called many times.

gamerson: Agreed, all in the constructor would be better.  I'll rework it.

gamerson: But the initialize() call is what is blocking the UI thread in some cases.  So we need to let the BndContainerInitializer go ahead and setup what it can (even though many classpath entries wont yet be realized because they need to be downloaded) and then once they are "available" then it can "reinit".

bjhargrave: I think you need two onWorkspaceInit calls. One in the constructor to add the listener and another here to do the project initialization.

gamerson:
I've tested this latest pull but I've still seen a few quirks with it. 

1) For very long running downloads of artifacts I've seen the repositories view go missing (have to close and reopen the view to get the elements back).
2) WorkspaceMainPart UI is not lazying loading correctly.

gamerson:
Regarding my previous message, I've fixed problem 2) with WorkspaceMainPart and I haven't seen the RepositoriesView problem.  I'm now back in the US and with faster download speeds I haven't see the same problem with problem 1)   

I'm at the point now that I think it would be good if this PR got some wider testing by bndtools users, maybe  I can make a post on the bndtools mailing list asking for volunteers

bndtools.core/src/bndtools/central/Central.java
```diff
@@ -212,6 +215,20 @@ public synchronized static Workspace getWorkspace() throws Exception {
...
+            for (Object plugin : newWorkspace.getPlugins()) {
+                if (plugin instanceof ProgressPlugin) {
+                    progressPlugins.add(plugin);
+                }
+            }
+
+            for (Object plugin : progressPlugins) {
+                newWorkspace.removeBasicPlugin(plugin);
+            }
+
```
bjhargrave: Shouldn't you just remove the BndProgress plugin vs. all ProgressPlugins?

gamerson: So in the httpClient it just calls, registry.getPlugin(ProgressPlugin.class) and I assume it would just grab the first one.  So I wanted to make sure JobProgress was the only one.

bjhargrave: Hmm, ok. Seems like it really should support multiple ProgressPlugins and not just one. But I guess this is ok for now.

gamerson: Should I create a bnd issue for that?

On Wed, Jun 22, 2016 at 2:05 PM, BJ Hargrave <notifications@github.com>
wrote:

> In bndtools.core/src/bndtools/central/Central.java
> <https://github.com/bndtools/bndtools/pull/1393#discussion_r68115321>:
>
> > @@ -212,6 +215,20 @@ public synchronized static Workspace getWorkspace() throws Exception {
> >              newWorkspace.addBasicPlugin(new WorkspaceListener(newWorkspace));
> >              newWorkspace.addBasicPlugin(instance.repoListenerTracker);
> >              newWorkspace.addBasicPlugin(getWorkspaceR5Repository());
> > +
> > +            // need to remove any progress plugins that already exist so we can favor JobProgress
> > +            Set<Object> progressPlugins = new HashSet<>();
> > +
> > +            for (Object plugin : newWorkspace.getPlugins()) {
> > +                if (plugin instanceof ProgressPlugin) {
> > +                    progressPlugins.add(plugin);
> > +                }
> > +            }
> > +
> > +            for (Object plugin : progressPlugins) {
> > +                newWorkspace.removeBasicPlugin(plugin);
> > +            }
> > +
>
> Hmm, ok. Seems like it really should support multiple ProgressPlugins and
> not just one. But I guess this is ok for now.
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/bndtools/bndtools/pull/1393/files/8c0bd9911f92fec87e4f0bea2b4d164121cfcf99..ef1896e7395a2dcf0e60ddb4e2300640816cd40d#r68115321>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAkVFcwOdCdbQwA01ORyj_qkAlBRtC5rks5qOYeDgaJpZM4IvtOy>
> .
>



-- 
Greg Amerson
Liferay Developer Tools
Liferay, Inc. www.liferay.com

bjhargrave: I guess ask @pkriens.

pkriens: Yes, I thought there was only one but then we need to support multiple.

gamerson: @pkriens  Do we need to have support for multiple ProgressPlugins in HttpClient before we merge this code?

pkriens: Yup, I think that is the clean solution?

gamerson: Done: https://github.com/bndtools/bnd/pull/1525

gamerson: Also I've removed the previous commit that removed the progress plugins, instead now its adds JobProgress since https://github.com/bndtools/bnd/pull/1525 adds support for multiple progress plugins

bndtools.core/src/bndtools/central/Central.java
```diff
@@ -212,22 +212,25 @@ public synchronized static Workspace getWorkspace() throws Exception {
...
-            workspace = newWorkspace;
-
-            workspaceRepositoryChangeDetector = new WorkspaceRepositoryChangeDetector(workspace);
+            workspaceRepositoryChangeDetector = new WorkspaceRepositoryChangeDetector(newWorkspace);
 
             // Call the queued workspace init callbacks
-            while (!workspaceInitCallbackQueue.isEmpty()) {
-                Function<Workspace,Void> callback = workspaceInitCallbackQueue.remove(0);
-                callback.run(workspace);
+            synchronized (workspaceInitCallbackQueue) {
```
bjhargrave: I am afraid of this new synchronization here. Especially since, while holding it, calls are made to the callbacks. I think it would be much safer to do something like is done in OSGi's PromiseImpl. Use a ConcurrentLinkedQueue to hold the call backs and a notifyCallbacks method which just returns if the workspace field is unset. This method can be called from both getWorkspace and onWorkspaceInit and it does not matter that multiple threads can be pulling callbacks off the queue and calling them. The Central class lock should probably not be be held either while calling the callbacks.

Just trying to avoid creating a future problem for ourselves :-) but I am open to this suggestion being overkill...

gamerson:
@bjhargrave beware, I'm writing the new sync code late at night, I could be _all wet_.

bndtools.core/src/bndtools/central/Central.java
```diff
@@ -239,11 +238,21 @@ public synchronized static Workspace getWorkspace() throws Exception {
...
-            callback.run(workspace);
-        else
-            workspaceInitCallbackQueue.add(callback);
+    public static void onWorkspaceInit(Function<Workspace,Void> callback) {
+        callbacks.add(callback);
+        notifyCallbacks();
+    }
+
+    private static void notifyCallbacks() {
+        synchronized (instance) {
```
bjhargrave: instance could be null if this method is called from onWorkspaceInit before Central is inited. I would suggest to remove the synchronized(instance) and just make the workspace field volatile.

bndtools.core/src/bndtools/central/Central.java
```diff
@@ -212,24 +212,23 @@ public synchronized static Workspace getWorkspace() throws Exception {
...
-
-            workspaceRepositoryChangeDetector = new WorkspaceRepositoryChangeDetector(workspace);
+            workspaceRepositoryChangeDetector = new WorkspaceRepositoryChangeDetector(newWorkspace);
 
-            // Call the queued workspace init callbacks
-            while (!workspaceInitCallbackQueue.isEmpty()) {
-                Function<Workspace,Void> callback = workspaceInitCallbackQueue.remove(0);
-                callback.run(workspace);
+            // The workspace has been initialized fully, set the field now
+            synchronized (instance) {
```
bjhargrave: Make workspace field volatile and remove synchronized(instance).

bndtools.core/src/bndtools/central/Central.java
```diff
@@ -212,24 +212,23 @@ public synchronized static Workspace getWorkspace() throws Exception {
...
+                workspace = newWorkspace;
             }
 
+            notifyCallbacks();
```
bjhargrave: The getWorkspace method is synchronized, so the call here is made holding the Central class monitor. It would be much better to rearrange to call this after the Central class monitor is released.

bndtools.core/src/bndtools/central/Central.java
```diff
@@ -239,11 +238,21 @@ public synchronized static Workspace getWorkspace() throws Exception {
...
+
+    private static void notifyCallbacks() {
+        synchronized (instance) {
+            if (workspace == null)
+                return;
+        }
+
+        while (!callbacks.isEmpty()) {
+            Function<Workspace,Void> callback = callbacks.poll();
+            callback.apply(workspace);
```
bjhargrave: You should wrap the apply call to catch exceptions:

```
try {
  callback.apply(workspace);
} catch (Throwable t) {
  //log the exception
}
```

bndtools.core/src/bndtools/central/Central.java
```diff
@@ -239,11 +238,21 @@ public synchronized static Workspace getWorkspace() throws Exception {
...
+        notifyCallbacks();
+    }
+
+    private static void notifyCallbacks() {
+        synchronized (instance) {
+            if (workspace == null)
+                return;
+        }
+
+        while (!callbacks.isEmpty()) {
```
bjhargrave: This loop is dangerous because you separate the isEmpty test from the poll call. This thread could see isEmpty is false, then another thread could remove the only item, then this thread could call poll, get null and then NPE at the apply call.

Safer is to do:

```
for (Function<Workspace,Void> callback = callbacks.poll(); callback != null; callback = callbacks.poll()) {
```

bndtools.core/src/bndtools/central/Central.java
pkriens: SHould we use a Promise here? This was a promise avant la lettre!

bjhargrave: Yes. We could use a Promise to hold the workspace object. Then onWorkspaceInit would just add the callback to the Promise and naturally call it when the Promise is resolved with the workspace object. This would be better than trying to replicate the Promise logic for callback notification.

gamerson:
@bjhargrave I made all the changes minus the Promise changes.  I have not written any Promise code heretofore, so I need to read up on them a bit, to make sure I understand their proper usage.

bjhargrave:
If you want, I can take what you have so far and make the Promise change. Let me know.

gamerson:
Sent to BJ here: https://github.com/bjhargrave/bndtools/pull/1  

closing this PR and let BJ take it across the finish line.

Issue #1374, opened 2016-05-18
New workspace wizard: does not respect chosen build plugin


fhuberts:
I can no longer generate a workspace with my plugin because the generation does not repect the chosen build plugin of the preferences.

It seems to be hard-coded to the one that is in the bndtools tree and that one generated files even when it is disabled in the preferences

major

fhuberts:
bndtools 3.2.0.REL

fhuberts:
it seems that commit a0059ef4dbaff8280f414ed34dcf8f65f003631a by @njbartlett  broke this

njbartlett:
I assume you're talking about the "Headless Build Systems" panel in the preferences? I missed that, it should have been removed.

Anything that goes into the workspace should now be generated by the user's selected template.

fhuberts:
That would completely kill support for my gradle plugin, which I'd obviously would not be very happy about. Was this done on purpose?

See also the other bug I filed: the legacy templates do not work.

So how is this supposed to work in the new setup?

njbartlett:
I don't know anything about your gradle plugin. I designed the new template system to make it far easier for anybody to design their own workspace and project templates. Did I do *that* on purpose? Of course.

The easiest way to create a workspace template is to create a GitHub project for it. See https://github.com/bndtools/workspace for an example.

njbartlett:
And yes I'll take a look at your other issue regarding the legacy templates.

fhuberts:
Please re-open

How are the gradle build files generated in this new system?
Like mentioned in the first comment: "It seems to be hard-coded to the one that is in the bndtools tree and that one generated files even when it is disabled in the preferences"

fhuberts:
I have users of my plugin, this bug kill functionality for my users.

I reported: "New workspace wizard: does not respect chosen build plugin" which is still very much true and the bug is therefore not fixed.
The report is not tied to my plugin, please do read the first text box of the report.

njbartlett:
I did read everything you wrote.

I don't understand what you believe to be "hard-coded". In the new template system, EVERYTHING in the workspace is generated by the template. So the template generates the gradle files, or the ant files, or the maven files or even Makefiles if it likes.

fhuberts:
Well, I just checked the code and what you did is remove support for HeadlessBuildPlugin plugins and VersionControlIgnoresPlugin plugins in the cnf setup. Everything else is still in place.

[...]

PR #1394, opened 2016-06-07
#1374: Proposed API -- DO NOT MERGE

njbartlett:
Proposed API for template post-processors


org.bndtools.templating/src/org/bndtools/templating/PostProcessor.java
```diff
@@ -0,0 +1,43 @@
...
+public interface PostProcessor {
+
+    /**
+     * <p>
+     * Service registration property indicating the kind(s) of templates this post-processor can target. Currently
+     * defined values are 'project', 'workspace' and 'bndrun'. The type of this property is String+. The post-processor
+     * shall be ignored if it does not specify at least one kind.
+     * </p>
+     */
+    public static String PROP_TEMPLATE_KIND = "kind";
```
fhuberts: why not use an array?
why not use an enum?

org.bndtools.templating/src/org/bndtools/templating/PostProcessor.java
```diff
@@ -0,0 +1,43 @@
...
+     *            The input parameters that were supplied to the template.
+     * @param resources
+     *            The resources generated by the template. A post-processor may add, remove or modify any entries in
+     *            this map.
+     * @param monitor
+     *            A monitor for progress reporting.
+     * @return
+     * @throws Exception
+     */
+    IStatus modifyOutputs(String templateKind, Template template, Map<String,List<Object>> templateParams, ResourceMap resources, IProgressMonitor monitor) throws Exception;
```
fhuberts: If at all possible I'd like this decoupled of Eclipse interfaces.
bndtools could provide internal conversion.

I gather that a plugin can modify the template and when I return bndtools will actually generate them on the filesystem?

bjhargrave:
@njbartlett What is going on with this? Should this be closed or is there progress to be made?

njbartlett:
Don’t close yet please, I need to pick this up again soon.

Issue #1399, opened 2016-06-13
WorkspaceSetupWizard not working for external Locations (restart required)


lostiniceland:
When using the WorkspaceSetupWizard to create a cnf-project at a external location (not within eclipse-workspace), new projects 

**Reproduce**
1. CRTL + N
2. BndTools -> Bnd OSGi Workspace
3. Set "Location" to some path outside the eclipse-workspace
4. Finish other steps (doesnt matter which template you choose)
5. Create New BndTools Project

![broken-workspace](https://cloud.githubusercontent.com/assets/1516116/16023582/a8903e8e-31bd-11e6-9eae-454ac7728238.png)
Project cannot be build, RepositoriesView not updated

**Reason**
[Central](https://github.com/bndtools/bndtools/blob/master/bndtools.core/src/bndtools/central/Central.java) is **caching a BND-workspace in a static field**. This field is initalized during startup with the eclipse-workspace-root if no cnf-project is found.
**The field is never updated**!

**Possible Solution**
Provide some reset-method in Central or listen on some project-modification (e.g. cnf created) which updates the field (and corresponding attributes).

A fix is not only needed to have a proper workspace-setup using external location. A fix is also needed for the [import-workspace-issue (916) ](https://github.com/bndtools/bndtools/issues/916) to reset Central after a cnf-project was imported.

lostiniceland:
Better solution is to hook into the ResourceChangeListener in order to encapsulate the reset-behavior. I am working on a solution...

lostiniceland:
This issue has been solved in 3.3.
I was not able to reproduce it in 3.3 RC1

Issue #916, opened 2014-09-30
Feature request: seamless project import into eclipse


eliasvasylenko:
It would be nice if Bndtools registered an import feature with eclipse, so that developers could check out from source control and start working straight away without having to first commit .project and .classpath files.

pkriens:
Neil, would this be something for you?

njbartlett:
Not sure I understand what is being asked for. It sounds like the existing UI provided by the eGit project.

gamerson:
Hey guys I think Elias is asking for supporting this use-case:

- developers creates brand-new bndtools style workspace (clone from
osgi/workspace, etc)
- developers creates a few bundle projects manually in the directory (maybe
copy/paste from somewhere?)
- now developer wants to open this bndtools workspace including all the
projects but no .classpath or .project files have ever been created.

So I believe Elias is asking for a import wizard that has the "smarts" to
add .project/.classpath files accordingly to those existing bundle projects
and then open those projects in eclipse.

On Tue, Mar 15, 2016 at 8:51 PM, Neil Bartlett <notifications@github.com>
wrote:

> Not sure I understand what is being asked for. It sounds like the existing
> UI provided by the eGit project.
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/bndtools/bndtools/issues/916#issuecomment-197106843>
>



-- 
Greg Amerson
Liferay Developer Tools
Liferay, Inc. www.liferay.com

njbartlett:
I'm going to close this because I don't think it's a common enough use-case... why not just create the projects in Bndtools?

I will consider a PR if somebody cares enough about this feature to implement it.

lostiniceland:
I am working on a PR for this (started on this several month ago but did not have the time to finish it). Simple import is working but UI and complex setups needs further improvement.

PR #1401, opened 2016-06-16
1399: WorkspaceSetupWizard not working for external Locations

lostiniceland:
Fixes: [1399](http://github.com/bndtools/bndtools/issues/issue/1399)

This PR also fixes several NullPointerExceptions during startup/shutdown when no bnd-workspace is available.

### DIFF
--- a/bndtools.core/src/bndtools/central/Central.java
+++ b/bndtools.core/src/bndtools/central/Central.java
@@ -20,10 +20,7 @@ import org.bndtools.api.ModelListener;
 import org.eclipse.core.resources.IFile;
 import org.eclipse.core.resources.IProject;
 import org.eclipse.core.resources.IResource;
-import org.eclipse.core.resources.IResourceChangeEvent;
 import org.eclipse.core.resources.IResourceChangeListener;
-import org.eclipse.core.resources.IResourceDelta;
-import org.eclipse.core.resources.IResourceDeltaVisitor;
 import org.eclipse.core.resources.IWorkspaceRoot;
 import org.eclipse.core.resources.ResourcesPlugin;
 import org.eclipse.core.runtime.CoreException;
@@ -53,6 +50,7 @@ import aQute.bnd.osgi.Constants;
 import aQute.bnd.osgi.Processor;
 import aQute.bnd.service.Refreshable;
 import aQute.bnd.service.RepositoryPlugin;
+import bndtools.Plugin;
 import bndtools.central.RepositoriesViewRefresher.RefreshModel;
 
 public class Central implements IStartupParticipant {
@@ -66,6 +64,7 @@ public class Central implements IStartupParticipant {
 
     static WorkspaceR5Repository r5Repository = null;
 
+    private static IResourceChangeListener resourceChangeListener;
     private static Auxiliary auxiliary;
 
     static final AtomicBoolean indexValid = new AtomicBoolean(false);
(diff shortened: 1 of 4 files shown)


bjhargrave:
@gamerson, @pkriens Can one or both of you review this change?

bndtools.core/src/bndtools/central/Central.java
```diff
@@ -204,39 +203,52 @@ public static Workspace getWorkspace() throws Exception {
...
                 return ws;
             }
-            try {
-                Workspace.setDriver(Constants.BNDDRIVER_ECLIPSE);
-                Workspace.addGestalt(Constants.GESTALT_INTERACTIVE, new Attrs());
+            ws = initializeWorkspace();
+            // The workspace has been initialized fully, set the field now
+            workspace = ws;
+        }
+        workspaceQueue.resolve(ws); // notify onWorkspaceInit callbacks
```
bjhargrave: workspaceQueue is resolved with the original Workspace. If you reset the workspace field to a new Workspace, you still end up with all callers to onWorkspaceInit being run against the original workspace. If you change workspace fields, you also need to change the workspaceQueue field to a new Deferred which is resolved with the new workspace. This needs to be done properly. The current code is built around these fields being written once and then never changed. If you plan to change them, then a much different strategy is needed to handle the invariant that they being changed together. The onWorkspaceInit static method also needs to be changes to handle that the workspaceQueue field can change over time.

bjhargrave:
This PR needs a lot of work to handle changing the current invariant that `workspace` is set only once and that `workspaceQueue` is final and resolved to the single `workspace` value. `onWorkspaceInit` also assumes that `workspaceQueue` is final and always set.

lostiniceland:
Thanks for the review. I will try to incorporate all remarks.

bndtools.core/src/bndtools/central/Central.java
lostiniceland: I had a look at the code again. As far as I understand the Promise, workspaceQueue is resolved with the new workspace instance, not the old one. This causes the callbacks registered in onWorkspaceInit to be fired with the same new instance.

Also workspaceQueue does not need to change over time, since it only holds Functions that have no relation to the workspace. Again, the workspace is passed during resolve.

The only case which was not handled for the callbacks is a null workspace. I've added a check that the callbacks will only be called if workspace is set.
The current (master) implementation cannot handle null workspaces...it will just end up in an exception which gets logged.

bndtools.core/src/bndtools/central/Central.java
```diff
@@ -655,4 +587,57 @@ public static void removeRepositoriesViewer(TreeViewer viewer) {
     public static void setRepositories(TreeViewer viewer, RefreshModel model) {
         repositoriesViewRefresher.setRepositories(viewer, model);
     }
+
+    public static void refreshRepositoriesViewers() {
+        repositoriesViewRefresher.updateViewers();
+    }
```
lostiniceland: Not sure if this is needed any more since the new onWorkspaceInit-callback solution

lostiniceland:
Somehow OSGIFrameworkContentProvider is broken (not related to this PR)

pkriens:
My bad ... fixed

bndtools.core/src/bndtools/central/Central.java
bjhargrave: > As far as I understand the Promise, workspaceQueue is resolved with the new workspace instance, not the old one.

workspaceQueue is a final variable and a Promise can be resolved exactly once. So once the promise is resolved with a workspace, all callbacks will use that workspace.


> EDIT: I've added an additional workspaceQueue.resolve in resetWorkspace

A second call to resolve will throw an IllegalArgumentException.

lostiniceland: Ok, back to the drawing board.

Unfortunately this PR was created before the Promise was added (just merged the changes). Any ideas/hints about how to relax the invariant?

bjhargrave: You cannot change the invariant that a Deferred/Promise can only be resolved once. You would need to replace the Deferred with a new one. And you need to make sure onWorkspaceInit work across any change. You will now have the workspace/workspaceQueue variables paired and thus you need a critical section to access either of them since they must change together. This then makes things much more complex. This is why changing the code from them being set only once to being reset is hard. Is there not a better solution to your issue which does not require them to be reset?

Also, there is a bug with the changes to add onWorkspaceInit which I have raised which may obviate the whole thing anyway :-) #1430.

lostiniceland: When I saw the changes to Central I was actually hoping that they render this PR obsolete :-)
I tested the SetupWorkspaceWizard on master but it didn´t work with an external location without restarting Eclipse. I will do another test since there have been some changes since the last time.

The main reason for this PR was issue #916 (I just found #1399 because I was looking for a working example)
I already have a working import ready, but it suffers the same problem as WorkspaceSetupWizard...you have to restart Eclipse to reset Central.

**If Central is not initialized with anything until there is a proper cnf project (in Eclipse), both issues should work without this PR.**

Waiting for #1430 is probably best. 
If you want to close this PR, feel free to do so. If necessary I will open a new one, assuming I found a proper solution with the workspace-invariant.

pkriens:
I think a lot of this stuff is different now and it should more or less work. There is now ALWAYS a workspace. When a cnf directory becomes available, we update that workspace and when it goes away we update it again. This means that all things should listen to their listeners.

lostiniceland:
Going to give my changes for issue 916 another try with the new workspace-handling. Due to family and job I didnt have any time to investigate this yet

lostiniceland:
I can confirm that this PR is not needed any more.

PR #1409, opened 2016-06-20
Update Documentation

lostiniceland:
Added section for changing bndlib to point to local file. Changed
structure to correct nesting.

Signed-off-by: Marc Schlegel <marc.schlegel@gmail.com>

### DIFF
--- a/DEV_README.md
+++ b/DEV_README.md
@@ -8,8 +8,8 @@ Bndtools aims to be a thin wrapper over bnd. This means that it is comparatively
 should be creating errors or warnings. In most cases bnd is responsible for generating errors at
 build time, it's then bndtools job to display them nicely to the user.
 
-Step 1 - Customise the bnd error
---------------------------------
+### Step 1 - Customise the bnd error
+
 
 Make sure that bnd is generating an appropriate error or warning using bnd's reporting API. By default
 all build errors are displayed as markers on the bnd.bnd file. This isn't usually the right file, and
@@ -26,8 +26,8 @@ e.g.
         .details(new org.bndtools.example.MyCustomizedLocationObject());
 
 
-Step 2 - Create a BuildErrorHandler
------------------------------------
+### Step 2 - Create a BuildErrorHandler
+
 
 Once bnd is generating extra error information then bndtools can use it to generate appropriate 
 markers. This is achieved through the use of the org.bndtools.build.api.BuildErrorDetailsHandler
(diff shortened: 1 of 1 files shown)


DEV_README.md
```diff
@@ -83,8 +82,30 @@ in to this you need to add the following to your plugin.xml
...
+Since BndTools relies on Bnd, it is sometimes necessary to use the local version of Bnd. This dependency is configured in `cnf/build.bnd` with a declaration of `bndlib` which is referenced in all projects.
+
+	bndlib:		aQute.libg;version='[${bnd-version-base},${bnd-version-ceiling})', \
+                biz.aQute.bndlib;version='[${bnd-version-base},${bnd-version-ceiling})'
+
+### Eclipse
+
+In order to configure Eclipse, change bndlib to reference the file generated in your Bnd-Workspace
+
+	bndlib		<YOUR SYSTEM PATH>/bnd/biz.aQute.bndlib/generated/biz.aQute.bndlib.jar;version=file
```
bjhargrave: This is wrong. You don't need to change this. All you need to is change `bnd_repourl` in gradle.properties. See https://github.com/bndtools/bndtools/blob/master/cnf/build.bnd#L15 and https://github.com/bndtools/bndtools/blob/master/cnf/ext/repositories.bnd#L14.

lostiniceland: I updated the PR, though I kept Peters solution as a alternative since you can skip the full release-build.

DEV_README.md
```diff
@@ -83,8 +82,32 @@ in to this you need to add the following to your plugin.xml
...
+
+The Repository for Bnd is configured with `bnd_repourl` in `gradle.properties`, located at the workspace-root. This file already contains a commented entry to point to a local distribution.
+
+	#bnd_repourl=../bnd/dist/bundles
+
+Assuming Bnd is located next to BndTools, just uncomment this line, otherwise adjust according to your needs, and release Bnd (see [documentation](https://github.com/bndtools/bnd) on how to release).
+
+Restart Eclipse or change some file in `cnf/ext` to pick-up the changes.
+
+### Alternative
```
bjhargrave: I don't like this alternative in here. It is only partial since it does not cover the other parts of the bnd build which are used by bndtools (embedded-repo, repository, etc.). Please remove this alternative section.

lostiniceland: Done.

I had some problems with the Gradle release: the release-task did not update the distribution when I didnt delete the folder manually. Hence I thought it would be good to have this documented as well. 
But if this is just a partial solution it´s definately better to skip this here.

Issue #1395, opened 2016-06-08
Source lookup fails in debugger


pkriens:
It seems that source lookup recently got worse, though not sure.

My main hangup is that I can add the sources to the code on the -buildpath (in OSGI-OPT/src). I can then set a breakpoint but the frigging debugger cannot find them. 

I tried to add the Bnd container (as Java Library) but this crashes. We probably have to look into to add this container to the source lookup.

@gamerson something for you? It is a real pain in the ass ...

gamerson:
@pkriens is the debugger stopping? but the Eclipse sourcecode lookup is not loading the editor correctly? Could you share a sample setup that is easy to reproduce and I can take a look.

lostiniceland:
Not sure if this is related, but I had similar problems with provider-bundles. I code against the API and the resolver add the providers to -runbundles. Those bundles are not part of the Bnd-Container in Eclipse so the degugger will not open any class even though when I add the provider-source (if available in workspace).

My current workaround, is to create a boot-project where all run-files are stored. Within this project I add all bundles that I want to debug to the buildpath. Now I can debug the providers.

I was thinking of a fix for this: before launching, BndTools will add all remaining bundles which are inside runbundles but not in the buildpath to the Bnd-Container. Once the running instance is stopped, the Bnd-Container is resetted.

gamerson:
The problem seems to be if the project whose sources you are trying to debug that you have open in Eclipse, if its .classpath file has a src folder that uses a non-default output location, for example, in the following classpath file the first classpathentry of `kind="src"` specifies its output location instead of leaving it default location (which means just omit the output attribute).

```
<?xml version="1.0" encoding="UTF-8"?>
<classpath>
	<classpathentry kind="src" output="bin" path="src"/>
	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-1.8"/>
	<classpathentry kind="con" path="aQute.bnd.classpath.container"/>
	<classpathentry kind="output" path="bin"/>
</classpath>
```

In this situation the  `DefaultProjectClasspathEntry` that is used by JavaRuntime when computing the classpath of a JUnit launches, when it looks at java project reference from the BndContainer and in that Java project it finds a it finds a `non-default` output location then it tries to add it to the classpath as a eclipse folder source container not a java project source container.  See here:

[...]

PR #1411, opened 2016-06-20
Improvement for drop on repositories view + test project for greg

pkriens:


bndtools.test/bndws/test.debug.dep/src/test/debug/dep/Foo.java
```diff
@@ -0,0 +1,7 @@
+package test.debug.dep;
+
+public class Foo {
+	{
+		System.out.println("Foo created");
```
gamerson: @pkriens  Is it here where you set the breakpoint and eclipse can't load it?

bndtools.test/bndws/test.debug/src/test/bottom/TopTest.java
```diff
@@ -0,0 +1,12 @@
+package test.bottom;
+
+import test.debug.dep.Foo;
+import junit.framework.TestCase;
+
+public class TopTest extends TestCase {
+
+	public void testX() {
+		new Foo();
+		System.out.println("hello");
```
gamerson: or here?

pkriens:
I set a breakpoint at setX and then single step. So testX is loaded correctly but then Foo.java is not editable. Thanks for working on this!

gamerson:
so Foo.java is loaded from OSGI-OPT/src and thus uneditable instead of being loaded from the real java project?

pkriens:
Yup … for some reason this has priority.

gamerson:
I remember when this change happened in 3.x dev.  It was during some improvements to "real classpath" visibility made in BND container awhile back where the generated jar was listed first on the classpath before the "eclipse project reference".  I recall this was on purpose but I don't remember the exact reason. 

cc @bjhargrave

pkriens:
Didn’t Neil do some work? Any chance we can get into the heart of the source selector? Would be great if we could skip the project on the container.

gamerson:
I've found the cause of this.  It is a little involved and has a few moving parts (project .classpath config and also bnd.bnd -buildpath config) so I will move that discussion to here https://github.com/bndtools/bndtools/issues/1395

gamerson:
@pkriens quick fix for you, in the test.debug.dep/.classpath, open that file, edit the first entry change it from

`<classpathentry kind="src" output="bin" path="src"/>`
to
`<classpathentry kind="src" path="src"/>`

Delete any existing JUnit launches and then relaunch JUnit.  That should make things work.

PR #1413, opened 2016-06-21
#1395 remove specific output location for main project source folders

### LINKED ISSUE #1395: Source lookup fails in debugger

gamerson:
fixes #1395


gamerson:
closing in favor of #1419

PR #1419, opened 2016-06-24
fix for Issue1395 with quick fix

gamerson:


gamerson:
In this PR I've add the marker resolution handlers.  This can be invoked from either the Problems/Markers view context menu or in the Project bnd editor.

screen shot 2016-06-24 at 1 45 34 am

screen shot 2016-06-24 at 1 47 31 am

bjhargrave:
So for test source folders which need a custom output location (e.g. bin_test), I guess they will just have poor source lookup when debugging tests?

gamerson:
The problem is dependent project source lookups.  So as long as the sources for tests are coming from the currrent projects /bin_test things are ok.  The problem would occur is sources for ProjectA/bin_test were coming from a dependent ProjectB/bin_test.  That is not nearly as common I would think.

bjhargrave:
OK, I am surprised M2E is not upset by this:

```
<classpathentry kind="src" output="target/classes" path="src/main/java">
```

gamerson:
Yes, that is interesting.  Do they have any other annotations on their classpath or perhaps their container is resolving those project entries in a different way that our BndContainer?

So there is another fix for this. It involves creating several extension points and classes around the JavaRuntimeClasspathResolver framework.  That way we can handle the custom output locations more gracefully, but that seemed like a lot of work.

gamerson:
I will debug m2e and see why the Eclipse java launch classpath entries are resolving correctly in that context.

gamerson:
@bjhargrave m2e does this
```
<extension point="org.eclipse.jdt.launching.classpathProviders">
      <classpathProvider 
          id="org.eclipse.m2e.launchconfig.classpathProvider"
          class="org.eclipse.m2e.jdt.internal.launch.MavenRuntimeClasspathProvider"/>
      <classpathProvider 
          id="org.eclipse.m2e.launchconfig.sourcepathProvider"
          class="org.eclipse.m2e.jdt.internal.launch.MavenSourcePathProvider"/>
   </extension>
```

So in their `MavenRuntimeClasspathProvider` they subclass the `StandardClasspathProvider` that bndtools is using.  Also they have this on startup:

```
this.launchConfigurationListener = new MavenLaunchConfigurationListener();
DebugPlugin.getDefault().getLaunchManager().addLaunchConfigurationListener(launchConfigurationListener);
```

So for *every* launch in Eclipse they check it to see if it is a `maven project` that is being `launched`.  If source they swap out the StandardClasspathProvider launch attribute for their own MavenRuntimeClasspathProvider.  In this class they are doing the right thing.

bjhargrave:
Yes! I thought there was something we could do. Can you look into this for bndtools? It is a real solution instead of just not using non-default output.

gamerson:
Sure, I'll take a stab.

Issue #1430, opened 2016-06-30
Delaying the population of the bnd classpath containers can cause workspace build failure


bjhargrave:
The change in BndContainerInitializer to delay population of the bnd classpath container, https://github.com/bndtools/bndtools/blob/5f5da3e1108f23b5919c314def69a234b2746136/bndtools.builder/src/org/bndtools/builder/classpath/BndContainerInitializer.java#L79-L92, causes problems in the OSGi build which builds plugins in the workspace.

![error](https://cloud.githubusercontent.com/assets/277682/16487815/50f91cc8-3ecd-11e6-825c-3399f8e2c398.jpg)

```
	at Packaging.<init>(Packaging.java:7)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
	at java.lang.Class.newInstance(Class.java:442)
	at aQute.bnd.osgi.Processor.loadPlugin(Processor.java:701)
	at aQute.bnd.osgi.Processor.loadPlugins(Processor.java:588)
	at aQute.bnd.osgi.Processor.getPlugins(Processor.java:477)
	at aQute.bnd.osgi.Processor.getPlugins(Processor.java:423)
	at aQute.bnd.build.Project.prepare(Project.java:348)
	at aQute.bnd.build.Project.getOutput(Project.java:872)
	at aQute.bnd.build.Project.getBundles(Project.java:580)
	at aQute.bnd.build.Project.parseBuildpath(Project.java:507)
	at aQute.bnd.build.Project.prepare(Project.java:366)
	at aQute.bnd.build.Project.getDependson(Project.java:792)
	at aQute.bnd.build.Workspace.getBuildOrder(Workspace.java:525)
	at bndtools.central.Central.getWorkspace(Central.java:219)
	at org.bndtools.builder.CnfWatcher.resourceChanged(CnfWatcher.java:40)
	at org.eclipse.core.internal.events.NotificationManager$1.run(NotificationManager.java:299)
	at org.eclipse.core.runtime.SafeRunner.run(SafeRunner.java:42)
[...]

PR #1432, opened 2016-06-30
container: Properly initialize bnd class path container when requested

bjhargrave:
Fixes https://github.com/bndtools/bndtools/issues/1430

### DIFF
--- a/bndtools.builder/src/org/bndtools/builder/classpath/BndContainerInitializer.java
+++ b/bndtools.builder/src/org/bndtools/builder/classpath/BndContainerInitializer.java
@@ -75,21 +75,10 @@ public class BndContainerInitializer extends ClasspathContainerInitializer imple
 
     @Override
     public void initialize(IPath containerPath, final IJavaProject javaProject) throws CoreException {
-        Updater.ensureClasspathContainer(javaProject); // We must ensure the container exists before this method returns
-        Central.onWorkspaceInit(new Function<Workspace,Void>() {
-            @Override
-            public Void apply(Workspace a) {
-                try {
-                    IProject project = javaProject.getProject();
+        IProject project = javaProject.getProject();
 
-                    Updater updater = new Updater(project, javaProject);
-                    updater.updateClasspathContainer(true);
-                } catch (CoreException e) {
-                    logger.logError("Error initializing classpath container", e);
-                }
-                return null;
-            }
-        });
+        Updater updater = new Updater(project, javaProject);
+        updater.updateClasspathContainer(true);
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


gamerson:
It seems this change will have the chance to block the UI.  I've seen cases where eclipse wants to initialize classpath containers on the UI thread (like if the classpath container is open in Package Explorer on restart of Eclipse).  

But I have see the problem that #1430 talks about.  So we need to address this problem.  What about this?  After BndContainer is properly initialized, what if we serialize the classpath entries to a IMemento.  Then on restart of Eclipse, if Eclipse needs to initialize the BndContainer early before Workspace is ready (and possibly could be on the UI thread) we could initially restore the BndContainer from the IMemento serialization.  So 99% of the time the container would be "correct enough" to not cause tons of errors seen by #1430 and then once Workspace is ready, we do the "real init" of BndContainer and throw away the serialized entries.  

What do you think?  That would prevent the UI blocking of initing the workspace.

bjhargrave:
> What do you think? That would prevent the UI blocking of initing the workspace.

Lets hold off for a little while. @pkriens has a replacement for FixedIndexRepo which will improve workspace startup time when there is no change to the index files. This is a source of time delay which, if ameliorated, will mean we really don't need to jump through a load of hoops to delay workspace init.

gamerson:
@bjhargrave even if Peter fixes FixedIndexRepo, the problem I mentioned would still remain with MavenBndRepository which is the type of repo I've done all of my UI testing with.

bjhargrave:
But I think it is fundamentally wrong to fake out the container until later. An IMemento approach is of little use if the file system changes outside of Eclipse (e.g. a git checkout of a different branch before starting Eclipse). The file system has the correct information, we need to use it to set the classpath container if Eclipse demands the classpath container be set on the UI thread. The better solution would be for Eclipse to not request to initialize the classpath container on the UI thread.

We should make bnd load projects as fast as possible to reduce the work needed to setup the classpath container but we should not set the classpath containers to false information.

bjhargrave:
> After BndContainer is properly initialized, what if we serialize the classpath entries to a IMemento. Then on restart of Eclipse, if Eclipse needs to initialize the BndContainer early before Workspace is ready (and possibly could be on the UI thread) we could initially restore the BndContainer from the IMemento that was serialized earlier (before restart of eclipse).

@gamerson How would we do this? I poked around the eclipse API and could not see how a memento for the container/project can be obtained from the arguments passed to `BndContainerInitializer.initialize(IPath, IJavaProject)`. Mementos seem to be associated with UI elements and the classpath container is not a UI element.

bjhargrave:
I am going to merge this for now. If we figure our a memento approach, we can make a new PR.

gamerson:
@bjhargrave I was incorrect I don't think there is a built-in way to use IMemento.  I was just thinking we could use it manually use the eclipse Memento APIs if we wanted to serialize so its human readable, m2e guys just write a binary state file to serialize their container.

PR #1443, opened 2016-07-22
[launching] add BndContainerRuntimeClasspathEntryResolver

gamerson:


gamerson:
@bjhargrave I did some more testing.  The bug I thought I had caused turned out to be reproducible with master branch (I'll add a separate bug)

gamerson:
Here is the other issue that is related to launches and source containers but it is not related to this new resolver. https://github.com/bndtools/bndtools/issues/1444

bjhargrave:
I am fine with you merging this in.

PR #1446, opened 2016-07-22
Propertly serialize BndDependencySourceContainer memento fixes #1444

### LINKED ISSUE #1444: Bnd OSGi Launch configuration will by itself remove Bnd Dependencies source container
The Bnd OSGi Launcher launch configuration when it is created initially has two source containers under the "source tab" in the launch configuration

- Default
- Bnd Dependencies

This is correct.  However, if you "edit" an existing launchconfiguration and you "expand" the Bnd Dependencies or the "Default" containers just to view them and then you try to exit the "launch configuration" dialog, Eclipse prompts you for "saving" the changes to the launch configuration.  However, it is unclear what "changes" have been made because we simply just expanded the "Dependencies" node in the source view.  

When you "save" the changes and then go back to that launch configuration the "Bnd Dependencies" source container is gone and then it breaks source lookup.   I've posted a screencap of this error:

http://g.recordit.co/MAHdgTPkjI.gif

gamerson:


bjhargrave:
LGTM

PR #1451, opened 2016-07-27
Attempt to get eclipsr to find sources of ajrs in external repositories

tomq42:
Previously the code appeared to try and convert the external file path into a workspace path
before creating an AcrhiveSourceContainer object.
This just, rather naively, directly constructs an ExternalArchiveSourceContainer pointing at the
bundle jar.

It appears to fix the issue of finding source for external libraries. If you have, say, downloaded
a bundle from Central, and added source, then source will now be found. Previously it appeared
that source was only found for bundles in Local, as only then was the jar file actually in the workspace.


tomq42:
Replace with another one

PR #1453, opened 2016-07-27
Same project put in debug source path multiple times

tomq42:
If there are multiple bundles in a bndrun configuration, and multiple of them
originate from the same project, that project is added to the debug source
search path multiple times. This commit just simply detects the duplication.

Signed-off-by: Tom Quarendon <tom.quarendon@teamwpc.co.uk>

### DIFF
--- a/bndtools.core/src/bndtools/launch/BndDependencySourceContainer.java
+++ b/bndtools.core/src/bndtools/launch/BndDependencySourceContainer.java
@@ -1,8 +1,10 @@
 package bndtools.launch;
 
 import java.util.Collection;
+import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Set;
 
 import org.bndtools.api.ILogger;
 import org.bndtools.api.Logger;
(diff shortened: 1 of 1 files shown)


gamerson:
LGTM

bndtools.core/src/bndtools/launch/BndDependencySourceContainer.java
```diff
@@ -65,13 +67,17 @@ public ISourceContainerType getType() {
...
         try {
             Project project = LaunchUtils.getBndProject(config);
             if (project != null) {
                 Collection<Container> runbundles = project.getRunbundles();
                 for (Container runbundle : runbundles) {
                     if (runbundle.getType() == TYPE.PROJECT) {
                         String targetProjName = runbundle.getProject().getName();
+                        if (projectsAdded.contains(targetProjName))
+                            continue;
+                        projectsAdded.add(targetProjName);
```
bjhargrave: This would be better as:
```java
if (projectsAdded.add(targetProjName)) {
   IProject targetProj = ...
}
```

That is use the boolean return value from `add` to decide whether to add the JavaProjectSourceContainer instead of a separate `contains` check.

tomq42:
Yes, good point.
Done.

PR #1455, opened 2016-07-27
Attempt to get eclipse to find sources of jars in external repositories

tomq42:
Previously the code appeared to try and convert the external file path into a workspace path
before creating an AcrhiveSourceContainer object.
This just, rather naively, directly constructs an ExternalArchiveSourceContainer pointing at the
bundle jar.

It appears to fix the issue of finding source for external libraries. If you have, say, downloaded
a bundle from Central, and added source, then source will now be found. Previously it appeared
that source was only found for bundles in Local, as only then was the jar file actually in the workspace.

Signed-off-by: Tom Quarendon <tom@quarendon.net>

### DIFF
--- a/bndtools.core/src/bndtools/launch/BndDependencySourceContainer.java
+++ b/bndtools.core/src/bndtools/launch/BndDependencySourceContainer.java
@@ -17,6 +17,7 @@ import org.eclipse.debug.core.sourcelookup.ISourceContainerType;
 import org.eclipse.debug.core.sourcelookup.ISourceLookupDirector;
 import org.eclipse.debug.core.sourcelookup.containers.ArchiveSourceContainer;
 import org.eclipse.debug.core.sourcelookup.containers.CompositeSourceContainer;
+import org.eclipse.debug.core.sourcelookup.containers.ExternalArchiveSourceContainer;
 import org.eclipse.jdt.core.IJavaProject;
 import org.eclipse.jdt.core.JavaCore;
 import org.eclipse.jdt.launching.sourcelookup.containers.JavaProjectSourceContainer;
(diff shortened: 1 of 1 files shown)


bndtools.core/src/bndtools/launch/BndDependencySourceContainer.java
```diff
@@ -78,14 +75,8 @@ public ISourceContainerType getType() {
...
                     } else if (runbundle.getType() == TYPE.REPO) {
-                        IPath bundlePath = Central.toPath(runbundle.getFile());
-                        if (bundlePath != null) {
-                            IFile bundleFile = ResourcesPlugin.getWorkspace().getRoot().getFile(bundlePath);
-                            if (bundleFile != null) {
-                                ArchiveSourceContainer tempArchiveCont = new ArchiveSourceContainer(bundleFile, false);
-                                result.add(tempArchiveCont);
-                            }
-                        }
+                        ExternalArchiveSourceContainer container = new ExternalArchiveSourceContainer(runbundle.getFile().toString(), false);
```
gamerson: So I believe if `IPath bundlePath` and `IFile bundleFile` is not null that means we have a workspace archive so we should use `ArchiveSourceContainer` as before.  But if those aren't in the workspace we could fallback to using the `ExternalArchiveContainer`.

tomq42:
As far as I can tell, the only thing that ArchiveSourceContainer actually does is report the name as a workspace path, rather than as an os path. All other operations are delegated to a ExternalArchiveContainer anyhow. So the only effect will be how the location is reported in the "edit source lookup path" dialog, or in the run configuration dialog. 
But certainly no problem in leaving the original path in as well. Probably more friendly to report a workspace path if there is one.

bjhargrave:
Please squash your commits and make sure the commit has the proper Signed-off-by. Thanks.

gamerson:
LGTM

tomq42:
OK, sorry.

PR #1457, opened 2016-07-27
simplfy category.xml feature tag and add description to category label

gamerson:


gamerson:
With this change the category label matches the feature description.

screen shot 2016-07-27 at 6 12 02 pm

PR #1460, opened 2016-07-29
Bnd Editor: Show correct DS selection in drop down

seanbright:
Because DS annotations are now the default, the drop down in the Bnd
editor is confusing. This fixes that and also changes the meaning of
"None" to mean disabled and update the model accordingly.

### DIFF
--- a/bndtools.core/src/bndtools/editor/contents/GeneralInfoPart.java
+++ b/bndtools.core/src/bndtools/editor/contents/GeneralInfoPart.java
@@ -88,7 +88,7 @@ public class GeneralInfoPart extends SectionPart implements PropertyChangeListen
     };
 
     private static enum ComponentChoice {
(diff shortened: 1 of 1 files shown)


bndtools.core/src/bndtools/editor/contents/GeneralInfoPart.java
```diff
@@ -111,6 +111,7 @@ public String getLabel() {
 
     private static final ServiceComponent SERVICE_COMPONENT_STAR = new ServiceComponent("*", new Attrs());
     private static final String DSANNOTATIONS_STAR = "*";
+    private static final String NO_DSANNOTATIONS_PATT = "!*";
```
bjhargrave: The no annotations should be empty string.

    -dsannotations:

is the way to disable ds annotations processing.

bjhargrave:
LGTM

PR #1465, opened 2016-08-04
[repository selector] clean repo naming in this view

rotty3000:


rotty3000:
The only issue here is with the Plugin type repos for which the naming is hard to obtain. I think this whole thing needs a cleanup, but this change minimally addresses the issue for now.

PR #1467, opened 2016-08-09
[resolution] include fragments in results

rotty3000:


rotty3000:
this is totally the wrong solution as root cause of the lost fragments was determined to be a bug in bnd https://github.com/bndtools/bnd/issues/1559

PR #1469, opened 2016-08-15
CellLabelProvider.getToolTipText() should return null instead of empty string.

seanbright:


bjhargrave:
LGTM

PR #1472, opened 2016-08-17
Use HttpClient during repository DnD

seanbright:
Java's built-in HttpUrlConnection will not follow redirects across
protocols (http -> https). bnd's HttpClient will, so use that instead
when dropping URLs on a repository.


bjhargrave:
LGTM

PR #1481, opened 2016-08-18
Move static methods to silence private reference warning

seanbright:
There is a lone problem marker in the Bndtools Eclipse workspace that reads:

`Export bndtools.central,  has 1,  private references [bndtools.model.repo]`

This is because RepositoryUtils exposed RepositoryBundle and
RepositoryBundleVersion as arguments to a couple static methods. This patch
moves those static methods to a new class in a non-exported package to
squelch the warning.

Signed-of-by: Sean Bright <sean.bright@gmail.com>

### DIFF
--- a/bndtools.core/src/bndtools/central/RepositoryUtils.java
+++ b/bndtools.core/src/bndtools/central/RepositoryUtils.java
@@ -4,25 +4,18 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.Callable;
+
 import org.bndtools.api.ILogger;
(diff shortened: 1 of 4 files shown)


seanbright:
Because I am removing methods from `bndtools.central` do I have to bump the package version to 2.0.0?

bjhargrave:
> Because I am removing methods from bndtools.central do I have to bump the package version to 2.0.0?

I don't think so personally. I am not sure who would be writing code dependent upon this package besides the code in this repo.

seanbright:
So good to merge then?

bjhargrave:
> So good to merge then?

I am ok with it.

PR #1483, opened 2016-08-18
[standalone] Seriously improved the support for -standalone

pkriens:
- Ordering and editing now works fine
- No more need to save when switching between standalone/workspace
- Organized code

Signed-off-by: Peter Kriens <peter.kriens@aqute.biz>


bndtools.core/src/bndtools/central/Central.java
```diff
@@ -228,6 +237,7 @@ public static Workspace getWorkspace() throws Exception {
                 if (ws != null) {
                     ws.close();
                 }
+                workspaceQueue.fail(e); // notify onWorkspaceInit callbacks
```
bjhargrave: You cannot resolve the deferred in the synchronized block since the callback will occur holding the lock.

PR #1492, opened 2016-08-19
defer processing of cnf changes until there is actually a workspace

gamerson:


bndtools.builder/src/org/bndtools/builder/CnfWatcher.java
```diff
@@ -33,16 +36,36 @@ static CnfWatcher install() {
...
     @Override
-    public void resourceChanged(IResourceChangeEvent event) {
+    public void resourceChanged(final IResourceChangeEvent event) {
+        if (Central.isWorkspaceInited()) {
+            processEvent(event);
+        } else {
+            Central.onWorkspaceInit(new Success<Workspace,Void>() {
+                @Override
+                public Promise<Void> call(Promise<Workspace> resolved) throws Exception {
+                    final Deferred<Void> completion = new Deferred<>();
```
bjhargrave: You don't need a Deferred here. Just return `Promises.resolved(null)` from the try and `Promises.failed(e)` from the catch.

PR #1499, opened 2016-08-23
various async cleanup and RepoView link for new workspaces

gamerson:


gamerson:
Here is the screenshot of the ReposView for a eclipse instance where there is no "cnf" folder.
screen shot 2016-08-23 at 12 01 14 pm

PR #1522, opened 2016-09-06
Git prefs improvements

seanbright:


org.bndtools.templating.gitrepo/src/org/bndtools/templating/jgit/GitRepoPreferences.java
```diff
@@ -8,10 +8,11 @@
 import org.osgi.framework.FrameworkUtil;
 
 import aQute.bnd.header.Parameters;
+import aQute.bnd.osgi.About;
 
 public class GitRepoPreferences {
 
-    public static final String INITIAL_GITHUB_REPOS = "osgi/workspace, bndtools/workspace;branch=origin/3.3";
+    public static final String INITIAL_GITHUB_REPOS = String.format("osgi/workspace, bndtools/workspace;branch=origin/%d.%d", About.CURRENT.getMajor(), About.CURRENT.getMinor());
```
bjhargrave: This is a bit risky. We basically keep the major.minor version of Bnd and Bndtools in sync, but if we don't do that one day for some reason, this could be a problem.

A solution would be to put the repo string in a resource which is -includeresource: {file} into the bundle. Then Bnd will macro process the repo string in the resource to use a macro (to be added) holding the major.minor version of Bndtools.

seanbright: The same could be said for the template repository and bndtools. Maybe we just hard code it and then update it when there is a reason to?

PR #1536, opened 2016-09-09
Enable Bndtools users to toggle Workspace offline mode

seanbright:
* Add toggle button to Repositories view
* Save workspace offline state in preferences
* Add workspace offline mode to main Bndtools preference page


bndtools.core/src/bndtools/preferences/WorkspaceOfflineChangeAdapter.java
```diff
@@ -0,0 +1,19 @@
...
+import org.eclipse.jface.util.IPropertyChangeListener;
+import org.eclipse.jface.util.PropertyChangeEvent;
+
+public abstract class WorkspaceOfflineChangeAdapter implements IPropertyChangeListener, WorkspaceOfflineChangeListener {
+    @Override
+    public final void propertyChange(PropertyChangeEvent event) {
+        if (!event.getProperty().equals(BndPreferences.PREF_WORKSPACE_OFFLINE)) {
+            return;
+        }
+        workspaceOfflineChanged((boolean) event.getNewValue());
```
bjhargrave: I am not sure how this works since it calls the method below which has an empty body!

bjhargrave: @rotty3000 points out you have an anonymous subclass below which override the following method. Perhaps then, this type should not implement the methods. It is already an abstract class.

seanbright: The `WorkspaceOfflineChangeListener ` had more than one method before. I've removed the default implementation.

bndtools.core/src/bndtools/views/repository/RepositoriesView.java
```diff
@@ -443,6 +460,31 @@ public IStatus runInWorkspace(IProgressMonitor monitor) throws CoreException {
         }.schedule();
     }
 
+    private void configureOfflineAction() {
+        Workspace workspace = Central.getWorkspaceIfPresent();
```
gamerson: I believe we need a 'Central.onInit()' style callback here.  this will block since this method calls getWorkspace will will download the internet if someone has a lot of jars in their repo.

seanbright: There is a call to `workspace.setOffline(value);` just after creating the `Workspace` object in `getWorkspace()`. Is that too late?

gamerson: Ok i see now, i didn't read the full code, and I see that this method is not called during the initial painting of the UI in createPartControl() so it wouldn't block, carry on then.

PR #1547, opened 2016-09-20
Update YourKit acknowledgement

seanbright:


bjhargrave:
You may also want to update the README file to include YourKit like we acknowledge CloudBees. https://github.com/bndtools/bndtools/blob/master/README.md#release

Issue #1412, opened 2016-06-20
Run-Descriptor: split information in two editor-tabs


lostiniceland:
It was discussed in [issue 1397](https://github.com/bndtools/bndtools/issues/1397) that the editor for the run-descriptor needs some cleanup.

The consesus was to move the sections _Repositories_ and _Runtime Properties_ to another tab, because the current editor got cluttered with a lot of information.
_Core Runtime_ should remain on the current tab because the selection is necessary for resolution.
Documentation must also be considered.

- [ ] Move _Repositories_ and _Runtime Properties_ to new tab
- [ ] Update documentation (for example [enRoute](http://enroute.osgi.org/qs/300-application.html))

lostiniceland:
Questions

How is the bundle-descriptor affected? I remember a statement on the user-group that the run-tab will be removed from bundle-descriptors, so the user always has to create a run-descriptor?

Which Milestone? Since it is a slightly bigger UI change, I would recommend to do this in 4.0.0 (especially because the documentation needs to be updated as well).

PR #1569, opened 2016-10-25
#1412 Signed-off-by: Chris Boyd <wodencafe@gmail.com>

wodencafe:
Hi guys, here is the PR for the code changes of issue #1412 
:)


bjhargrave:
@gamerson Can you look at this to review from a Bndtools UI perspective?

gamerson:
hey @wodencafe 

Just a few comments from a UI review standpoint.

1. For each new page, we don't need the first `section` we can just host the controls directly on the page.  The section header is redundant since that are each on their own pages now
2. IMHO, we don't really need those form images for the Runtime/Repos page.  however, if you really would like to see them added would it make sense to have the "Repositories" image look more similar to the Repositories view image.

Also, can you squash all of your commits into one new commit and add a descriptive commit message with what changes you made?  Thanks!

wodencafe:
Hi guys, ok I will work on the changes this weekend and submit them :)

PR #1596, opened 2016-12-01
Clarify exception message in LaunchUtils

### LINKED ISSUE #1591: Exception after renaming project referenced from launch configuration
Steps:
1. create a launch configuration for an existing run descriptor
2. rename the project that contains the run descriptor
3. launch the launch configuration

The following error message is shown:

> Error initialising bnd launcher java.lang.IllegalArgumentException: Bnd launch target was not specified, or does not exist.

I Found the message difficult to understand. Wouldn't something like 
> The run descriptor 'org.example-project/foo.bndrun' could not be found.

be clearer?

BTW, having _Yes_ and _No_ buttons makes not sense here. Irrespective of which button is pressed, the launch configuration dialog is opened.

image

Stacktrace
```
java.lang.IllegalArgumentException: Bnd launch target was not specified, or does not exist.
	at bndtools.launch.util.LaunchUtils.getBndProject(LaunchUtils.java:52)
	at bndtools.launch.AbstractOSGiLaunchDelegate.buildForLaunch(AbstractOSGiLaunchDelegate.java:52)
	at org.eclipse.debug.internal.core.LaunchConfiguration.launch(LaunchConfiguration.java:855)
	at org.eclipse.debug.internal.core.LaunchConfiguration.launch(LaunchConfiguration.java:739)
	at org.eclipse.debug.internal.ui.DebugUIPlugin.buildAndLaunch(DebugUIPlugin.java:1039)
	at org.eclipse.debug.internal.ui.DebugUIPlugin$8.run(DebugUIPlugin.java:1256)
	at org.eclipse.core.internal.jobs.Worker.run(Worker.java:55)
```

rherrmann:
Fixes #1591: Exception after renaming project referenced from launch configuration

Signed-off-by: Rüdiger Herrmann <ruediger.herrmann@gmx.de>

### DIFF
--- a/bndtools.core/src/bndtools/launch/util/LaunchUtils.java
+++ b/bndtools.core/src/bndtools/launch/util/LaunchUtils.java
@@ -19,10 +19,7 @@ public final class LaunchUtils {
     private LaunchUtils() {}
 
     public static IResource getTargetResource(ILaunchConfiguration configuration) throws CoreException {
-        String target = configuration.getAttribute(LaunchConstants.ATTR_LAUNCH_TARGET, (String) null);
-        if (target == null || target.length() == 0) {
-            return null;
-        }
+        String target = getTargetName(configuration);
 
         IResource targetResource = ResourcesPlugin.getWorkspace().getRoot().findMember(target);
         if (targetResource == null)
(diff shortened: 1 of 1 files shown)


bndtools.core/src/bndtools/launch/util/LaunchUtils.java
```diff
@@ -19,10 +19,7 @@
     private LaunchUtils() {}
 
     public static IResource getTargetResource(ILaunchConfiguration configuration) throws CoreException {
-        String target = configuration.getAttribute(LaunchConstants.ATTR_LAUNCH_TARGET, (String) null);
-        if (target == null || target.length() == 0) {
-            return null;
-        }
+        String target = getTargetName(configuration);
 
         IResource targetResource = ResourcesPlugin.getWorkspace().getRoot().findMember(target);
```
bjhargrave: Is it OK to possible pass `null` to `findMember` here? You removed the null check that returns, so null can now be passed to `findMember`.

rherrmann: Of course it is not OK to pass `null`. Thanks for the hint. I will provide an updated pull request.

PR #1601, opened 2016-12-04
Remove unnecessary null check in LaunchUtils

### LINKED ISSUE #1600: Unnecessary null check in LaunchUtils::getTargetResource()
See https://github.com/bndtools/bndtools/blob/master/bndtools.core/src/bndtools/launch/util/LaunchUtils.java#L28

I think, the null check
```java
if (targetResource == null)
    return null;
```
is unnecessary.

Would you agree to remove the extra null check and inline the above line like this?
```java
return ResourcesPlugin.getWorkspace().getRoot().findMember(target);
```

rherrmann:
Closes #1600 Unnecessary null check in LaunchUtils::getTargetResource()
Signed-off-by: Rüdiger Herrmann <ruediger.herrmann@gmx.de>

### DIFF
--- a/bndtools.core/src/bndtools/launch/util/LaunchUtils.java
+++ b/bndtools.core/src/bndtools/launch/util/LaunchUtils.java
@@ -24,9 +24,6 @@ public final class LaunchUtils {
             return null;
 
         IResource targetResource = ResourcesPlugin.getWorkspace().getRoot().findMember(target);
-        if (targetResource == null)
-            return null;
-
         return targetResource;
     }
 


bndtools.core/src/bndtools/launch/util/LaunchUtils.java
```diff
@@ -20,13 +20,8 @@ private LaunchUtils() {}
 
     public static IResource getTargetResource(ILaunchConfiguration configuration) throws CoreException {
         String target = getTargetName(configuration);
-        if (target == null)
```
bjhargrave: This null check is still necessary since you do not want to pass null to findMember.

rherrmann: Sorry again, this wasn't intended at all.

PR #1641, opened 2017-01-28
launch: Add project name to launch configuration

jonpasski:
Newly created launch configurations do not include a project
name via an expected attribute, which can cause and endless
loop on launching under certain circumstances.

Fixes: 1639

Signed-off-by: Jon Passki <jon@passki.us>


gamerson:
LGTM

Issue #1625, opened 2017-01-20
NullPointerException while closing BndEditor


rherrmann:
Occasionally, I see a NullPointerException written to the Eclipse log. Looking into the code reveals, that the runnable (`BndEditor$4.run`) does not check if the objects it is accessing aren't disposed of yet.

Error log entry:
```
org.eclipse.swt.SWTException: Failed to execute runnable (java.lang.NullPointerException)
	at org.eclipse.swt.SWT.error(SWT.java:4441)
	at org.eclipse.swt.SWT.error(SWT.java:4356)
	at org.eclipse.swt.widgets.Synchronizer.runAsyncMessages(Synchronizer.java:139)
	at org.eclipse.swt.widgets.Display.runAsyncMessages(Display.java:4147)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:3764)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine$9.run(PartRenderingEngine.java:1151)
	at org.eclipse.core.databinding.observable.Realm.runWithDefault(Realm.java:332)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine.run(PartRenderingEngine.java:1032)
	at org.eclipse.e4.ui.internal.workbench.E4Workbench.createAndRunUI(E4Workbench.java:148)
	at org.eclipse.ui.internal.Workbench$5.run(Workbench.java:636)
	at org.eclipse.core.databinding.observable.Realm.runWithDefault(Realm.java:332)
	at org.eclipse.ui.internal.Workbench.createAndRunWorkbench(Workbench.java:579)
	at org.eclipse.ui.PlatformUI.createAndRunWorkbench(PlatformUI.java:150)
	at org.eclipse.ui.internal.ide.application.IDEApplication.start(IDEApplication.java:135)
	at org.eclipse.equinox.internal.app.EclipseAppHandle.run(EclipseAppHandle.java:196)
	at bndtools.eclipse.applaunch.BndApplicationLauncher$1.run(BndApplicationLauncher.java:37)
	at aQute.launcher.Launcher$5.call(Launcher.java:1214)
	at aQute.launcher.Launcher$5.call(Launcher.java:1212)
	at aQute.launcher.Launcher.run(Launcher.java:316)
	at aQute.launcher.Launcher.main(Launcher.java:133)
Caused by: java.lang.NullPointerException
[...]

PR #1643, opened 2017-02-05
Check for null when using IDocumentProvider #1625

wodencafe:
This ensures that IDocumentProvider is checked for null when used in BndEditor, as this item can be null if the form is disposed.


wodencafe:
Hi @bjhargrave , I have updated my commit message to include a signed off by tag :)

bjhargrave:
@wodencafe Your [commit here](https://github.com/wodencafe/bndtools/commit/825d3ab8aec4102cf745fb5b0c94edb2d49a4736.patch) has the same issue with bad email address as I mention in #1644. Please fix. Thanks.

wodencafe:
Hi @bjhargrave , I have updated my git config and amended the commit.

Issue #1598, opened 2016-12-01
Preference pages under Bndtools don't show when filtering for 'bnd'


rherrmann:
After entering 'bnd' in the filter input field of the Preferences dialog, only the main _Bndtools_ page is shown. Sub-pages like _Build_, _Generated Resources_, etc. are hidden

![image](https://cloud.githubusercontent.com/assets/607182/20799523/bef714bc-b7e2-11e6-8f0a-1f3d8ffbf2e5.png)

PR #1672, opened 2017-03-11
Preference pages under Bndtools don't show when filtering for 'bnd'

lostiniceland:
1598: Preference pages under Bndtools don't show when filtering for 'bnd'

fixes bndtools/bndtools#1598

Extensions-Points enhanced with keywords to include subpages in filter. Though Bndtools preferences are already at the root, showing subpages on filter is a expected behaviour in Eclipse.

Signed-off-by: Marc Schlegel maschlegel@gmail.com

### DIFF
--- a/bndtools.core/_plugin.xml
+++ b/bndtools.core/_plugin.xml
@@ -689,6 +689,10 @@
       </shortcut>
 
    </extension>
+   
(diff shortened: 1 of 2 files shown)


gamerson:
LGTM

Issue #1669, opened 2017-03-05
Repositories view should reuse standard refresh action


rherrmann:
The _Refresh_ action in the _Repositories_ should reuse the icon of the workbench refresh action (
![image](https://cloud.githubusercontent.com/assets/607182/23589713/57310272-01d2-11e7-8286-03d89a0f93c9.png)) and also react on the <kbd>F5</kbd> key binding like other refresh actions do.

PR #1681, opened 2017-04-08
Reuse Eclipse standard refresh action (and F5 support) #1669

wodencafe:
This addresses issue #1669, by using the Eclipse standard refresh icon, and adds F5 refresh support to the Repositories view.


bndtools.core/src/bndtools/views/repository/RepositoriesView.java
```diff
@@ -464,6 +466,17 @@ public IStatus runInWorkspace(IProgressMonitor monitor) throws CoreException {
                 return Status.OK_STATUS;
             }
         }.schedule();
+        KeyAdapter adapter = new KeyAdapter() {
```
gamerson: Refresh is one of those globally recognized key sequences in Eclipse.  So in this view we can register an action (that is a button on view toolbar) that hooks into the global 'refresh' action.  That way if the user customizes their refresh action globally, this will pick up on those changes instead of hard-coding to 'F5'.  I will try to find an example of this and send you the source.

gamerson: Here is a Eclipsepedia artifact about this. 
https://wiki.eclipse.org/FAQ_How_do_I_hook_into_global_actions,_such_as_Copy_and_Delete%3F

So you would want to do something like:
```
IActionBars actionBars= getViewSite().getActionBars();
   actionBars.setGlobalActionHandler(
      ActionFactory.REFRESH.getId(),
      refreshAction);	
```

wodencafe:
Hi @gamerson, thanks for the helpful info, good to know!

I have made the requested change and verified that the F5 support is working using **IActionBars**.

PR #1685, opened 2017-04-12
conditional package

gamerson:


bndtools.m2e/bnd.bnd
```diff
@@ -30,9 +30,9 @@ Import-Package: \
 	org.codehaus.plexus.util.xml;provider=m2e,\
 	*
 
-Private-Package: \
-	aQute.lib.collections;-split-package:=merge-first,\
-	bndtools.m2e.*
+Conditional-Package: aQute.lib*
```
bjhargrave: You may want: 

`Conditional-Package: aQute.lib*;-split-package:=first`

The `-split-package` directive will avoid any warnings if the package appears multiple times on the buildpath.

PR #1692, opened 2017-04-24
[workingset] Adds working sets to the bnd file.

pkriens:
You can now define working sets in the bnd.bnd file (or in build.bnd). There are 2 modes:

      -workingset: bar, foo, baz

This will add the project to the given working sets as one can expect.

The other mode is intended to be used in bnd (but not limited to)

      -workingset: Implementations; member=${filter;${p};.*\.impl\..*}

The member attribute is evaluated to a boolean with Processor.isTrue. If true, the project is a member of the corresponding working set. 

The code will create working sets as demanded but will reuse existing working set with the matching name. If no -workingset property is given, the working sets are not touched in any way for that project.

Signed-off-by: Peter Kriens <peter.kriens@aqute.biz>

### DIFF
--- a/bndtools.builder/src/org/bndtools/builder/WorkingSetTracker.java
+++ b/bndtools.builder/src/org/bndtools/builder/WorkingSetTracker.java
@@ -1,0 +1,104 @@
+package org.bndtools.builder;
+
+import java.util.Iterator;
+import java.util.List;
+
+import org.eclipse.core.resources.IProject;
+import org.eclipse.core.runtime.IAdaptable;
+import org.eclipse.swt.widgets.Display;
+import org.eclipse.ui.IWorkbench;
+import org.eclipse.ui.IWorkingSet;
+import org.eclipse.ui.IWorkingSetManager;
+import org.eclipse.ui.PlatformUI;
+
+import aQute.bnd.build.Project;
+import aQute.bnd.header.Attrs;
+import aQute.bnd.header.Parameters;
+import aQute.bnd.osgi.Processor;
+import aQute.lib.collections.ExtList;
+
+/**
+ * Utility to track working sets from the bnd.bnd file You can set the -workingset to a list of working set names. This
+ * will automatically add this project to that working set. If there is no such working set it will create one. For each
+ * parameter in the -workingset proeprty you can set an attribute {@code member}. If this attribute is truthy (as
+ * defined in Processor.isTrue) or not set then the project is made a member of the working set, otherwise it is
+ * removed.
+ *
+ * <pre>
+ * -workingset: foo
+ * -workingset: impl;member=${;${p};.*.impl.*}
+ * </pre>
+ */
+public class WorkingSetTracker {
+
+    static void doWorkingSets(final Project model, final IProject targetProject) {
+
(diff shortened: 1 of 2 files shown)


bndtools.builder/src/org/bndtools/builder/WorkingSetTracker.java
```diff
@@ -0,0 +1,104 @@
...
+                members.add(targetProject);
+            }
+
+            if (elements.length != members.size()) {
+                updateWorkingSet(currentWorkingSet, members);
+            }
+        }
+
+        for (final String name : memberShips.keySet()) {
+            Display.getDefault().asyncExec(new Runnable() {
```
rherrmann: `asyncExec` is not necessary here. The WorkingSetManager may be called from a non-UI thread.

bndtools.builder/src/org/bndtools/builder/WorkingSetTracker.java
```diff
@@ -0,0 +1,104 @@
...
+                    newWorkingSet.setLabel(null);
+                    workingSetManager.addWorkingSet(newWorkingSet);
+                }
+
+            });
+        }
+    }
+
+    static private void updateWorkingSet(final IWorkingSet wset, final List<IAdaptable> members) {
+        Display.getDefault().asyncExec(new Runnable() {
```
rherrmann: `asyncExec` is not necessary here. As far as I can see, the WorkingSetManager may be called from a non-UI thread.

bndtools.builder/src/org/bndtools/builder/WorkingSetTracker.java
```diff
@@ -0,0 +1,104 @@
...
+
+        for (final String name : memberShips.keySet()) {
+            Display.getDefault().asyncExec(new Runnable() {
+
+                @Override
+                public void run() {
+                    IAdaptable[] members = new IAdaptable[1];
+                    members[0] = targetProject;
+                    IWorkingSet newWorkingSet = workingSetManager.createWorkingSet(name, members);
+                    newWorkingSet.setId("org.eclipse.ui.resourceWorkingSetPage");
```
rherrmann: I suggest creating Java working sets (id is `org.eclipse.jdt.ui.JavaWorkingSetPage`) as these are more common among JDT users.

bndtools.builder/src/org/bndtools/builder/WorkingSetTracker.java
pkriens: How can one know if a manager can be called from any thread in Eclipse?

rherrmann: Some _managers_ document this detail, this one doesn't. Here I tested the code after removing the surrounding `asyncExec()` and it still worked as expected.

rherrmann: See this test case for an example: https://github.com/rherrmann/eclipse-extras/blob/master/com.codeaffine.extras.workingset.test/src/com/codeaffine/extras/workingset/internal/DynamicWorkingSetUpdaterPDETest.java#L228

PR #1706, opened 2017-05-19
There is nothing remarkable about a Refreshable returning false.

seanbright:


bjhargrave:
What is the point of this? Seems better to leave things as they are. Then if the Refreshable gets better, the code is ready.

seanbright:
If your repository list contains a `MavenBndRepository` you get an error spammed to the error log you do a refresh if nothing is actually updated. Currently, every in-tree `Refreshable` with the exception of `MavenBndRepository` always returns true or throws an Exception.

bjhargrave:
ok

PR #1725, opened 2017-06-07
[Baseline] Properly report Bundle-Version location

psoreide:
If Bundle-Version is defined in build.bnd or ext, no markers are created and the project will not build

Signed-off-by: PK Søreide <per.kristian.soreide@gmail.com>


psoreide:
Closing. Need to figure out how to get rid of the Markers from cnf.

PR #1726, opened 2017-06-07
[Baseline] Properly report Bundle-Version location

psoreide:
If Bundle-Version is defined in build.bnd or ext, no markers are created and the project will not build

The marker will be set on the Projects bnd.bnd file

Signed-off-by: PK Søreide <per.kristian.soreide@gmail.com>

### DIFF
--- a/bndtools.builder/src/org/bndtools/builder/handlers/baseline/BundleVersionErrorHandler.java
+++ b/bndtools.builder/src/org/bndtools/builder/handlers/baseline/BundleVersionErrorHandler.java
@@ -27,6 +27,7 @@ import aQute.bnd.differ.Baseline.BundleInfo;
 import aQute.bnd.osgi.Builder;
 import aQute.lib.io.IO;
 import aQute.service.reporter.Report.Location;
+import bndtools.central.Central;
 
 public class BundleVersionErrorHandler extends AbstractBuildErrorDetailsHandler {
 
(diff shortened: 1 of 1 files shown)


bndtools.builder/src/org/bndtools/builder/handlers/baseline/BundleVersionErrorHandler.java
```diff
@@ -23,17 +23,27 @@
...
 import aQute.service.reporter.Report.Location;
+import bndtools.central.Central;
 
 public class BundleVersionErrorHandler extends AbstractBuildErrorDetailsHandler {
 
     private static final String PROP_SUGGESTED_VERSION = "suggestedVersion";
 
     private final static String VERSION_ACCEPTING_MACRO_STRING = "(\\d+)\\.(\\d+)\\.(\\d+)\\.([-_\\.\\$\\{\\}\\da-zA-Z]+)";//$NON-NLS-1$
     private final static Pattern VERSION_ACCEPTING_MACRO = Pattern.compile(VERSION_ACCEPTING_MACRO_STRING);
+    private final static LineLocation DUMMY_LINE_LOCATION;
```
bjhargrave: I don't think it is safe to use a static here. Some other code could modify the LineLocation since it is not immutable. You must create a new LineLocation each time you need it.

bndtools.builder/src/org/bndtools/builder/handlers/baseline/BundleVersionErrorHandler.java
```diff
@@ -60,6 +70,36 @@
...
+                    IFile buildFile = Central.getWorkspaceBuildFile();
+                    loc = findBundleVersionHeader(buildFile);
+                    if (loc != null) {
+                        loc = DUMMY_LINE_LOCATION;
+                    }
+                }
+
+                if (loc == null) {
+                    // Not found in build.bnd, try ext folder. Marker will appear on bnd.bnd
+                    File extDir = new File(Central.getWorkspace().getBuildDir(), Workspace.EXT);
```
bjhargrave: You don't need to search the ext dir and it is also not complete. Just look at the files in `getWorkspace().getIncluded()`.

PR #1742, opened 2017-07-05
MavenProject can still be null when refreshing Maven metadata

timothyjward:
When trying out the latest code I experienced NPEs from the Bnd M2E builder. It happens when I force a Maven project update offline:

    java.lang.NullPointerException
	    at bndtools.m2e.BndConfigurator$1$1.runInWorkspace(BndConfigurator.java:61)
	    at org.eclipse.core.internal.resources.InternalWorkspaceJob.run(InternalWorkspaceJob.java:39)
	    at org.eclipse.core.internal.jobs.Worker.run(Worker.java:55)

I'm not an Eclipse expert, but it seems as though we should always be using the monitored version of getMavenProject() if the unmonitored one returns null...


timothyjward:
Note that the NPEs don't actually seem to prevent the build from eventually succeeding.

Issue #1765, opened 2017-09-07
Template repositories don't support authentication


brampouwelse:
Template repositories (configured in Eclipse preferences) dont work if they require autentication.

brampouwelse:
I had a look at the code to see if I can fix this one myself... And I did get it to work but discovered some other things. 

Loading templates doesn't work at all when using a repository that requires authentication
 - For template repositories (Eclipse preferences) templates are not listed at all in the dialog
 - For workspace repositories  (OSGiRepository) templates are listed but can't be loaded 

That it doesn't work for OSGiRepositories has to do with the `OSGiRepository` not supporting the way the `RepoPluginsBundleLocator` queries it. Eventually the `RepoPluginsBundleLocator` reverts to a direct download (which fails if the repository requires authentication);

bjhargrave:
OSGiRepositories use the http code which supports credentials. See http://bnd.bndtools.org/instructions/connection-settings.html for how to set them up.

brampouwelse:
I'm aware of how to use connection settings (which also work for a `FixedIndexedRepo` btw.) but it just doesn't work. 

The template repos that are configured in the Eclipse preferences are loaded using a FixedIndexedRepo that's not part of the workspace ( https://github.com/bndtools/bndtools/blob/master/bndtools.core/src/org/bndtools/core/templating/repobased/ReposTemplateLoader.java#L184 )

That repo just doesn't get the connection settings .. 

I've tried to come up with a fix ( https://github.com/bndtools/bndtools/pull/1766 )
but then ran into another issue. 

https://github.com/bndtools/bndtools/blob/master/bndtools.core/src/org/bndtools/core/templating/repobased/RepoPluginsBundleLocator.java#L28

The `RepoPluginsBundleLocator` tries to load the bundle from the repo without version, that doesn't work for the `OSGiRepository` so it always falls back to a direct download without beeing aware of the connection settings. This makes that templates are listed in the selection dialog but after that it can't download the bundle containing the template. 

The only way I managed to load templates from an authenticated repo at this moment is by adding a `FixedIndexedRepo` to the workspace (and configuring the connection settings).

PR #1766, opened 2017-09-07
#1765 Fix loading templates from authenticated repositories

brampouwelse:
Use an OSGiRepository instead of a FixedIndexedRepo for template repositories configured in the Eclipse preferences. 

Use the HttpClient when falling back to direct download in RepoPluginsBundleLocator. 

Signed-off-by: Bram Pouwelse <bram@pouwelse.com>

### DIFF
--- a/bndtools.core/src/org/bndtools/core/templating/repobased/ReposTemplateLoader.java
+++ b/bndtools.core/src/org/bndtools/core/templating/repobased/ReposTemplateLoader.java
@@ -1,10 +1,9 @@
 package org.bndtools.core.templating.repobased;
 
-import java.io.IOException;
-import java.net.URISyntaxException;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


bndtools.core/src/org/bndtools/core/templating/repobased/ReposTemplateLoader.java
```diff
@@ -95,7 +96,7 @@ void activate() {
             if (workspace == null)
                 workspace = Central.getWorkspace();
             workspaceRepos = workspace.getPlugins(Repository.class);
-            tmpLocator = new RepoPluginsBundleLocator(workspace.getRepositories());
+            tmpLocator = new RepoPluginsBundleLocator(workspace.getPlugins(RepositoryPlugin.class), workspace.getPlugin(HttpClient.class));
```
brampouwelse: I think it would be better to delay creation of the `RepoPluginsBundleLocator` until the preference configured repo is created and include that as well. 

But with the current `RepoPluginsBundleLocator` implementation that doesn't add any value as it seems not to work at all for `OSGiRepository`

bndtools.core/src/org/bndtools/core/templating/repobased/ReposTemplateLoader.java
```diff
@@ -172,23 +173,30 @@ private static void addPreferenceConfiguredRepos(List<Repository> repos, Reporte
...
         for (Iterator<String> iter = uris.iterator(); iter.hasNext();) {
             sb.append(iter.next());
             if (iter.hasNext())
                 sb.append(',');
         }
-        repo.setLocations(sb.toString());
+
+        Map<String,String> properties = new HashMap<>();
+        properties.put("name", "bndtools-preferences-template-repos");
+        properties.put("locations", sb.toString());
```
brampouwelse: I wasn't completely sure here .. might be safer to also add `max.stale=0` to prevent caching as caching for this repo can't be configured?

bjhargrave:
Change to use OSGiRepository done by 606fad0f52d7488b7d21dfd195ed670e755638a9

PR #1767, opened 2017-09-14
M2E support for the indexer plugin

timothyjward:
The `bnd-indexer-maven-plugin` is used to generate indexes when using maven. In turn those indexes are used in standalone bndruns when resolving/testing/exporting.

At the moment the only way to get the index to be regenerated is to run a build manually. This is easy to forget, and often means that the index metadata is out of sync with projects in the workspace. The first commit in this PR adds m2e build support for the indexer, including an automatic trigger when any of the indexed workspace project artifacts change.

The second commit in this PR updates the standalone bndrun editor to automatically reload the bnd model when the index changes. This means that the UI and the resolve button always work with up-to-date data after a build of the index, and avoids the user having to close/reopen the bndrun file.

This PR represents a significant usability improvement for Maven users and I would like for it to make it into the Bndtools 3.5.0 release.

### DIFF
--- a/bndtools.core/src/bndtools/editor/project/RepositoriesEditModel.java
+++ b/bndtools.core/src/bndtools/editor/project/RepositoriesEditModel.java
@@ -204,7 +204,7 @@ class RepositoriesEditModel {
         return r;
     }
 
-    private void updateStandaloneWorkspace(BndEditModel model) throws Exception {
+    void updateStandaloneWorkspace(BndEditModel model) throws Exception {
 
         assert isStandalone();
 
--- a/bndtools.core/src/bndtools/editor/project/RepositorySelectionPart.java
+++ b/bndtools.core/src/bndtools/editor/project/RepositorySelectionPart.java
@@ -1,15 +1,26 @@
 package bndtools.editor.project;
 
+import java.io.File;
 import java.lang.reflect.InvocationTargetException;
 import java.net.URI;
+import java.net.URISyntaxException;
 import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 
 import org.bndtools.core.ui.icons.Icons;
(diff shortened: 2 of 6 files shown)


njbartlett:
It kind-of works. I have found the following problems so far:

1. Nothing happens on projects where the indexer plugin is inherited. That is, I have the `bnd-indexer-maven-plugin` on my parent project, and so all the child projects inherit it. The indexes on these child projects are not regenerated. However if I explicitly declare the indexer plugin on a project then the index is regenerated.

2. If I have the bndrun file open in the editor when I make a change to a bundle, I notice that the editor goes into DIRTY state (i.e. unsaved changes). If I click Resolve, the new index doesn't seem to be used. However when I save the editor and then click Resolve, I get the new resolution results.

Aside from this, looking great! Thanks.

timothyjward:
> Nothing happens on projects where the indexer plugin is inherited. That is, I have the bnd-indexer-maven-plugin on my parent project, and so all the child projects inherit it. The indexes on these child projects are not regenerated. However if I explicitly declare the indexer plugin on a project then the index is regenerated.

I wasn't able to reproduce this exactly, however I did find an issue with the build not recognising that the index needed to be rebuilt if the project depended on its own output artifact. When I tried your example workspace with the updated PR commit it seemed to work correctly.

> If I have the bndrun file open in the editor when I make a change to a bundle, I notice that the editor goes into DIRTY state (i.e. unsaved changes). If I click Resolve, the new index doesn't seem to be used. However when I save the editor and then click Resolve, I get the new resolution results.

I have not been able to reproduce this at all, even when I have a different editor page open...

njbartlett:
Both these issues now seem to be fixed (I also cannot reproduce the second issue with the dirty editor state). Looks great! I do notice quite a lot of activity in the Progress view, I feel this needs testing on a larger scale project.

njbartlett:
Have done a bit more testing and this is really looking good. I'm going to merge in order to get wider feedback on the feature.

PR #1768, opened 2017-09-14
Enable ui decorators when using Maven to build

timothyjward:
The Bndtools builder adds useful decorations to exported packages, and to files which declare OSGi components. This should be exended to include the maven build as well.

Signed-off-by: Tim Ward <timothyjward@apache.org>


timothyjward:
Rebased on master to allow easier testing combined with other maven enhancements

njbartlett:
This looks good. I have tested that decorators are correctly removed when their root cause (e.g. a `@Component` or `@Version` annotation) is removed. NB issue #1774 affects this code branch, but that bug also affects non-Maven workspaces.

Issue #1780, opened 2017-09-26
Eclipse plugin/RCP development difficult without proper Target Definition


eliasvasylenko:
Developing e4 plug-ins and RCP apps is surprisingly great with Bndtools. The main missing piece imo is that a lot of discoverability in the e4 model editors depends on plug-ins being visible through the target definition. I assume the story with Eclipse 3.x plug-ins is similar.

Currently if we want workspace projects to see each other through the PDE tooling it seems we have to manually add generated/[project].jar to the target platform for every individual project (or various other hacky workarounds), which is an annoying maintenance burden. If we have external dependencies through repository types other than p2 it is even more of a pain to get them into the target definition.

The whole target definition concept seems a bit odd from the perspective of a Bnd developer rather than just resolving against the plug-ins on the project classpath, but this is the system we have to work with.

I propose contributing to the `org.eclipse.pde.ui.targetLocationProvisioners` and `org.eclipse.pde.core.targetLocations` extension points to allow adding Bnd repositories to the available target locations.

So if I work on this, would `bndtools.core` be the place for it? That seems to be where similar things are kept.

timothyjward:
> I propose contributing to the org.eclipse.pde.ui.targetLocationProvisioners and org.eclipse.pde.core.targetLocations extension points to allow adding Bnd repositories to the available target locations.

:+1: to having better support for RCP development. I don't know enough about PDE to say for certain that this is the way to go, but it seems to be worth prototyping.

> So if I work on this, would `bndtools.core` be the place for it? That seems to be where similar things are kept.

I would suggest that this probably belongs in a new project, particularly if it adds PDE dependencies. We may even want to make it an optional feature (like the bndtools M2E support).

eliasvasylenko:
Yes that makes sense, there were tons of new dependencies when I got started so I've taken your advice and moved into a new bndtools.pde project for now.

eliasvasylenko:
Any thoughts or suggestions?

timothyjward:
Thanks for your patience. There was a fair bit of work to get 3.5.0 out of the door, but now that's done it's a really good time to start working further on this. I've put some comments on the PR - I think it's mostly there, but there are a couple of issues.

filippor:
You can look at https://github.com/filippor/bndPdeTarget
Its a pde extension that can add bnd artifact to target definition

PR #1782, opened 2017-09-28
For #1780 Added bndtools.pde project for plugin development using Bndtools

eliasvasylenko:
For #1780 

Might be a bit rough around the edges, but the only major things missing I think are icon decorations for borked target locations, and a listener on file moves to track run descriptors. Anyway I'll be away for about a week so I wanted to submit this for review in the meantime.

The new project contains two target location extensions for exposing Bnd
workspaces to PDE tooling:

- RepositoryTargetLocation allows adding the contents of Bnd
repositories to target definitions

- RunDescriptorTargetLocation allows adding -runbundles from a run
descriptor to a target definition.

I'm also open to suggestions for further Bnd/PDE integration ...

Signed-off-by: Elias N Vasylenko <eliasvasylenko@gmail.com>


timothyjward:
I'm not a PDE expert, but it broadly looks fine to me from a code perspective. 

The only questions that I have are:

* Doesn't something need to be added to the p2 feature.xml to get the bundle installed?
* Should this be added as a separate feature, rather than an "always on" component? The M2E integration has https://github.com/bndtools/bndtools/blob/master/build/build.gradle#L170-L176 and https://github.com/bndtools/bndtools/blob/master/build/feature/bndtools-m2e/feature.xml to do this

It's probably best to do a rebase and force push now that we've moved to 4.0.0.DEV. A local build/install should also prove whether I'm correct about needing to update the feature definitions.

timothyjward:
I've tried using this locally, and I get some issues when trying to do a test run. One of them is that the new plugin doesn't resolve. I've included a couple of review comments that should fix that bit.

I've also confirmed that the bndtools p2 repository that gets built doesn't install the new bndtools.pde code.

eliasvasylenko:
Ah you're absolutely right, you even mentioned these things as I was getting started and I still ended up missing it! Oops. Thank you for taking a look at it, I'll make these changes as soon as I get a chance.

eliasvasylenko:
CI hasn't finished yet but it appears to work now, including installing the new PDE feature locally into an existing eclipse installation. I just copied the m2e stuff for license information attached to the feature so I assume it's all kosher.

timothyjward:
> CI hasn't finished yet but it appears to work now, including installing the new PDE feature locally into an existing eclipse installation. I just copied the m2e stuff for license information attached to the feature so I assume it's all kosher.

I can confirm that the install works, and that PDE integration appears as a selectable option. The installation didn't cause errors with either of my big workspaces so I'm happy for it to be merged. I don't do PDE development but it may be worth @peterkir taking a look at what (if anything) else could be added on top of this.

PR #1783, opened 2017-09-29
Avoid queuing up unnecessary re-indexing jobs

timothyjward:
Now that re-indexing is triggered on every dependency change we do a lot more checking of the Maven projects. These checks were done individually, but this turns out to take a long time. This change attempts to reduce the number of check jobs, aggregating the events where possible. I have seen a significant improvement (6 mins -> 16 seconds) as a result - it would be good if others could check this helps them too.

### DIFF
--- a/bndtools.m2e/src/bndtools/m2e/IndexConfigurator.java
+++ b/bndtools.m2e/src/bndtools/m2e/IndexConfigurator.java
@@ -108,6 +108,91 @@ public class IndexConfigurator extends AbstractProjectConfigurator implements IR
         }
     }
 
+    private final class RebuildIndexCheck extends WorkspaceJob {
+
+        private final List<IResourceChangeEvent> events = new ArrayList<>();
+        private final IMavenProjectFacade facade;
+
+        private boolean noMoreEvents;
+
+        public RebuildIndexCheck(String name, IResourceChangeEvent event, IMavenProjectFacade facade) {
+            super(name);
+            this.events.add(event);
+            this.facade = facade;
+        }
+
+        public void addEvent(IResourceChangeEvent event) {
+            synchronized (pendingJobs) {
+                if (!noMoreEvents) {
+                    events.add(event);
+                } else {
+                    throw new IllegalStateException("This job is underway, no new events can be added");
+                }
+            }
+        }
+
+        @Override
+        public IStatus runInWorkspace(IProgressMonitor monitor) throws CoreException {
+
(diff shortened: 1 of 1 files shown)


bndtools.m2e/src/bndtools/m2e/IndexConfigurator.java
```diff
@@ -108,6 +108,91 @@ public WorkspaceRepository getRepository() {
...
+        private boolean noMoreEvents;
+
+        public RebuildIndexCheck(String name, IResourceChangeEvent event, IMavenProjectFacade facade) {
+            super(name);
+            this.events.add(event);
+            this.facade = facade;
+        }
+
+        public void addEvent(IResourceChangeEvent event) {
+            synchronized (pendingJobs) {
```
bjhargrave: You can avoid synchronizing here since `addEvent` is only called when already synchronized on `pendingJobs`. This methods should also not be public: default access is safer.

bndtools.m2e/src/bndtools/m2e/IndexConfigurator.java
```diff
@@ -108,6 +108,91 @@ public WorkspaceRepository getRepository() {
...
+                if (!noMoreEvents) {
```
bjhargrave: The `noMoreEvent` boolean is not necessary. `addEvent` is only called under `pendingJob` lock when then job is in the map.

njbartlett:
I have tested on the problematic workspace, and this is definitely a massive improvement.

bndtools.m2e/src/bndtools/m2e/IndexConfigurator.java
timothyjward: I'd quite like to leave this in as it makes it much harder to accidentally use the Job incorrectly. Events really can't be added after a certain point in the RebuildIndexCheck lifecycle, and if someone tries to add one later then they really should get an exception. The other option is to add lots of JavaDoc explaining that you always have to synchronize and check the pendingJobs, but that seems like an accident waiting to happen...

bndtools.m2e/src/bndtools/m2e/IndexConfigurator.java
timothyjward: I agree with using default access, and I have realised that the inner class can be static. I'd like to keep the synchronized block, as it will continue to protect this method correctly if/when the IndexConfigurator is next changed/refactored.

PR #1801, opened 2017-11-16
 Support alias requirements in Run Reqs/Blacklist

njbartlett:


njbartlett:
Waiting on bndtools/bnd#2204

PR #1823, opened 2018-01-25
Templating: Work around NPE causing template loader failures

seanbright:

### DIFF
--- a/bndtools.core/src/org/bndtools/core/templating/repobased/ReposTemplateLoader.java
+++ b/bndtools.core/src/org/bndtools/core/templating/repobased/ReposTemplateLoader.java
@@ -121,6 +121,9 @@ public class ReposTemplateLoader implements TemplateLoader {
         // Map List<Repository> to Promise<List<Template>>
         Promise<List<Template>> promise = repos.stream().map(repo -> promiseFactory.submit(() -> {
             Map<Requirement,Collection<Capability>> providerMap = repo.findProviders(Collections.singleton(requirement));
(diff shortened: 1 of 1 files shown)


seanbright:
I'm sure there are some fancy Promise-y things that could be done here instead, but I defer to the reviewers.

bndtools.core/src/org/bndtools/core/templating/repobased/ReposTemplateLoader.java
```diff
@@ -121,6 +121,9 @@ void dectivate() {
         // Map List<Repository> to Promise<List<Template>>
         Promise<List<Template>> promise = repos.stream().map(repo -> promiseFactory.submit(() -> {
             Map<Requirement,Collection<Capability>> providerMap = repo.findProviders(Collections.singleton(requirement));
+            if (!providerMap.containsKey(requirement)) {
```
bjhargrave: This is not possible. Per the specification of the Repository.findProviders method:

```If there are no matching capabilities for a specified requirement, then the value in the map for the specified requirement must be an empty collection.```

So `requirement` MUST be a key in the returned map.

What is the NPE stacktrace?

seanbright: So I was using the current bndtools/workspace template (3.5) and the InfoRepositoryWrapper around the JPM repo was causing the problem. So this may be a bnd issue. I am looking now.

PR #1828, opened 2018-02-07
Fix a few minor problems during release

seanbright:
Two problems are resolved by this changeset:

1. When ProjectBuilder goes out of scope it closes all of the JARs in its close list and classpath. When we try to iterate those and get metadata, we get an exception because the JARs are already closed. Instead grab the metadata earlier and save it for later.
2. If a baselining error was present before an 'Update and Release' (i.e. "The bundle version (X.X.X/X.X.X) is too low...") we would throw up a dialog indicating that there was an error during the release when there was none. Instead we now clear the error list after the version update.

### DIFF
--- a/bndtools.release/src/bndtools/release/api/ReleaseContext.java
+++ b/bndtools.release/src/bndtools/release/api/ReleaseContext.java
@@ -32,7 +32,7 @@ public class ReleaseContext {
     private IProgressMonitor progressMonitor;
 
     private final List<Jar> releasedJars;
-    private final Map<String,Object> properties;
+    private final Map<String, Object> properties;
     private final ErrorHandler errorHandler;
     private Scope currentScope;
     private ReleaseOption releaseOption;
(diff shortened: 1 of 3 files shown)


bndtools.release/src/bndtools/release/api/ReleaseContext.java
```diff
@@ -75,13 +77,23 @@ public void setProgressMonitor(IProgressMonitor progressMonitor) {
...
         return releasedJars;
     }
 
+    public List<String> getReleaseSummaries() {
+        return releasedJarSummaries;
+    }
+
     public void close() {
+        // XXX: This code doesn't do anything, these JARs were closed by
+        // ProjectBuilder
```
bjhargrave: OK, so remove the close loop and leave a comment that indicates the released jars were closed by the project builder.

seanbright: K

bjhargrave:
@seanbright what is this PR waiting on to be merged?

PR #1842, opened 2018-03-04
[repoindex] replace repoindex with bnd simple indexer

rotty3000:


rotty3000:
Once this is in I can push the complete removal of repoindex (and bindex) from the bnd repo.

build/main.bnd
```diff
@@ -2,9 +2,10 @@
 -resourceonly: true
 -includeresource: \
     ${repo;biz.aQute.bndlib;[${bnd-version-base},${bnd-version-ceiling});HIGHEST},\
-    ${repo;org.osgi.impl.bundle.repoindex.lib;[${bnd-version-base},${bnd-version-ceiling});HIGHEST},\
     ${repo;biz.aQute.repository;[${bnd-version-base},${bnd-version-ceiling});HIGHEST},\
     ${repo;biz.aQute.resolve;[${bnd-version-base},${bnd-version-ceiling});HIGHEST},\
+    ${repo;org.apache.servicemix.bundles.kxml2;latest},\
```
bjhargrave: What added the requirement for these 2 bundles at runtime? They were not needed before. I know the packages were part of the remove bindex jar and grepping shows they are only used by some felix metatype code. 

So instead of adding a runtime dependency on 2 bundles, lets just Conditional-Package the kxml and xmlpull packages in the bundles which actually need them (which appears to be just bndtools.core).

rotty3000:
@bjhargrave I made the change you requested.

PR #1845, opened 2018-03-07
[git repo] fix compile error when obr is finally gone

rotty3000:


bndtools.builder/bnd.bnd
```diff
@@ -41,12 +41,9 @@ Import-Package: *;ui.workbench=!;common=!;registry=!;texteditor=!;text=!
 
 # Content
 repoindex-packages: \
-	org.osgi.service.indexer.impl.types,\
-	org.osgi.service.indexer.impl,\
-	org.osgi.service.indexer.impl.util,\
 	org.kxml2.io;-split-package:=merge-first,\
```
bjhargrave: I am wondering, why does `bndtools.builder` even need these 2 packages as Private-Packages? git grep shows they are not used by `bndtools.builder`. Can you not get rid of `repoindex-packages` all together here?

rotty3000: I think it's based on another dependency that is being slurped in... namely the apache felix metatype processing.

bjhargrave: But that class is in `bndtools.core` which has the private copy of these classes. I looked at `bndtools.builder` and I don't know why those packages where ever in there. No code in the bundle uses those packages per `bnd print`.

```
$ bnd print -b bndtools.builder/generated/bndtools.builder.jar 

[USEDBY]

org.xmlpull.v1                           org.kxml2.io
                           
```

rotty3000: kk, I will remove and see what happens.

rotty3000:
@bjhargrave updated as requested!

Issue #1408, opened 2016-06-19
add an import suggestion for packages from repositories


rotty3000:
It would be fantastic if bndtools could provide import suggestions from exports available in the repositories which would result in a `-buildpath` addition.

lostiniceland:
It sounds like a very nice feature, but I suspect it would be very easy to end up with a cyclic dependency, since all (exported) packages would be visible in code-completion.

kriegfrj:
Is this a duplicate of #1589?

Issue #1589, opened 2016-11-29
Make importing packages easy


rherrmann:
Consider this scenario: First, create this interface in bundle `org.example.api`:
```java
package org.example.api;

public interface Greeting {
  String sayHello( String name );
}
```
and exported the `org.example.api` package. Then create a service that implements the above interface in bundle `org.example.impl`:

```java
package org.example.impl;

import org.osgi.service.component.annotations.Component;

@Component
public class GreetingComponent implements Greeting {
  public String sayHello( String name ) {
    return "Hello " + name;
  }
}
```
Of course, the type `Greeting` cannot be resolved because the `org.example.api` package is not yet _known_ to the `impl` bundle. The bnd file of the `impl` bundle needs to be edited so that its `-buildpath` directive contains the `api` bundle.

I suggest to provide a quick fix that adds the missing bundle to the build path of the `impl` bundle.

kriegfrj:
I'm trying to port a large-ish PDE project to Bndtools, and this feature would save quite a bit of time in the process of adding lots of bundles to the build path. Is anyone working on it? If not, I can have a crack.

rotty3000:
I don't think anyone is working on this. ☺

PR #1850, opened 2018-03-14
core: Added quick fixes for missing bundles (implements #1589)

kriegfrj:
This PR is an implementation of #1589.

If you have an import statement for a package/class that is not on the build path, and Bndtools can find one or more bundles that exports that package in the known set of local repositories, then it will add a "quick fixes" to the top of the available quick fixes. Similarly for fully-qualified references to unknown types. 

Bundles from the current workspace will be listed first, then from other repositories.

If you select the "quick fix" for a given bundle, then that bundle will be added to the build path in the `bnd.bnd` file for the current project.

There are some heuristics in there to try to distinguish between class names and package names where it is syntactically ambiguous (basically, in most cases it is assumed to be a package name if it starts with a lower case letter and a class name otherwise).

I tried to have a fairly comprehensive set of automated tests as well, though I know I fell short of 100% coverage.

Possible future enhancements:

# Demote suggestions to add a bundle if a different bundle is already on the build path that exports the same package.
# Add auto-completion proposals for types/imports of packages from known bundles.
# Index the exported classes in the bundles as well as the exported packages, so that we could provide type fixes/completions as well as package fixes/completions.

### DIFF
--- a/bndtools.core/bnd.bnd
+++ b/bndtools.core/bnd.bnd
@@ -106,4 +106,7 @@ eclipse.deps: \
 -testpath: \
 	slf4j.api,\
 	slf4j.simple,\
+	org.assertj.core,\
+	org.mockito.mockito-core,\
+	org.objenesis,\
 	${junit}
--- a/bndtools.core/src/org/bndtools/core/editors/ImportPackageQuickFixProcessor.java
+++ b/bndtools.core/src/org/bndtools/core/editors/ImportPackageQuickFixProcessor.java
@@ -1,0 +1,542 @@
+package org.bndtools.core.editors;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.io.UnsupportedEncodingException;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import org.bndtools.api.BndtoolsConstants;
+import org.bndtools.api.ILogger;
+import org.bndtools.api.Logger;
+import org.bndtools.core.ui.icons.Icons;
+import org.eclipse.core.resources.IFile;
+import org.eclipse.core.resources.IProject;
+import org.eclipse.core.resources.IResource;
+import org.eclipse.core.resources.IResourceRuleFactory;
+import org.eclipse.core.resources.IWorkspace;
+import org.eclipse.core.resources.ResourcesPlugin;
+import org.eclipse.core.resources.WorkspaceJob;
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.core.runtime.IStatus;
+import org.eclipse.core.runtime.Status;
+import org.eclipse.jdt.core.ICompilationUnit;
+import org.eclipse.jdt.core.IJavaProject;
+import org.eclipse.jdt.core.compiler.IProblem;
+import org.eclipse.jdt.core.dom.ASTNode;
+import org.eclipse.jdt.core.dom.ImportDeclaration;
+import org.eclipse.jdt.core.dom.Name;
+import org.eclipse.jdt.core.dom.NameQualifiedType;
+import org.eclipse.jdt.core.dom.PackageDeclaration;
+import org.eclipse.jdt.core.dom.QualifiedName;
+import org.eclipse.jdt.core.dom.SimpleType;
+import org.eclipse.jdt.internal.corext.dom.ASTNodes;
+import org.eclipse.jdt.ui.text.java.IInvocationContext;
+import org.eclipse.jdt.ui.text.java.IJavaCompletionProposal;
+import org.eclipse.jdt.ui.text.java.IProblemLocation;
+import org.eclipse.jdt.ui.text.java.IQuickFixProcessor;
+import org.eclipse.jface.text.contentassist.IContextInformation;
+import org.eclipse.swt.graphics.Image;
+import org.eclipse.swt.graphics.Point;
+import org.osgi.framework.namespace.BundleNamespace;
+import org.osgi.framework.namespace.IdentityNamespace;
+import org.osgi.framework.namespace.PackageNamespace;
+import org.osgi.resource.Capability;
+import org.osgi.resource.Requirement;
+import org.osgi.resource.Resource;
+import org.osgi.service.repository.Repository;
+
+import aQute.bnd.build.Project;
+import aQute.bnd.build.Workspace;
+import aQute.bnd.build.WorkspaceRepository;
+import aQute.bnd.build.model.BndEditModel;
+import aQute.bnd.build.model.clauses.VersionedClause;
+import aQute.bnd.header.Attrs;
+import aQute.bnd.osgi.resource.CapReqBuilder;
+import aQute.bnd.properties.Document;
+import aQute.bnd.properties.IDocument;
+import aQute.bnd.service.RepositoryPlugin;
+import aQute.lib.collections.MultiMap;
+import aQute.lib.io.ByteBufferInputStream;
+import aQute.lib.io.IO;
+import aQute.lib.strings.Strings;
+import bndtools.central.Central;
+import bndtools.central.RepositoryUtils;
+
+public class ImportPackageQuickFixProcessor implements IQuickFixProcessor {
+
+    private static final ILogger logger = Logger.getLogger(ImportPackageQuickFixProcessor.class);
+
+    // Relevance constants
+    public static final int ADD_BUNDLE = 15;
+    public static final int ADD_BUNDLE_WORKSPACE = ADD_BUNDLE + 1;
+
+    @Override
+    public boolean hasCorrections(ICompilationUnit unit, int problemId) {
+
+        switch (problemId) {
+            case IProblem.ImportNotFound :
+            case IProblem.UndefinedType :
+                return true;
+            default :
+                return false;
+        }
+    }
+
+    private static Requirement buildRequirement(String importName) {
+        final CapReqBuilder b = new CapReqBuilder(PackageNamespace.PACKAGE_NAMESPACE);
+        b.addFilter(PackageNamespace.PACKAGE_NAMESPACE, importName, null, new Attrs());
+        return b.buildSyntheticRequirement();
+    }
+
+    private Collection<Capability> searchRepository(Repository osgiRepo, String pkg) {
+        final Requirement req = buildRequirement(pkg);
+        Map<Requirement, Collection<Capability>> providers = osgiRepo.findProviders(Collections.singleton(req));
+        return providers.get(req);
+    }
+
+    static class BndBuildPathHandler {
+        private final IInvocationContext context;
+        private IFile bndFile;
+        private IDocument bndDoc;
+        private BndEditModel bndModel;
+        private List<VersionedClause> buildPath;
+        private Set<String> buildPathBundles;
+        private Set<VersionedClause> buildPathVersioned;
+        private IProgressMonitor monitor;
+
+        public BndBuildPathHandler(IInvocationContext context) {
+            this.context = context;
+        }
+
+        public IProgressMonitor getProgressMonitor() {
+            return monitor;
+        }
+
+        public void setProgressMonitor(IProgressMonitor monitor) {
+            this.monitor = monitor;
+        }
+
+        // This has been pushed into this method to defer initialization as this makes testing easier.
+        void loadFileInfo() {
+            if (bndFile != null) {
+                return;
+            }
+            final ICompilationUnit compUnit = context.getCompilationUnit();
+            final IJavaProject java = compUnit.getJavaProject();
+            final IProject eclipse = java.getProject();
+            bndFile = eclipse.getFile(Project.BNDFILE);
+        }
+
+        Workspace getWorkspace() throws Exception {
+            return Central.getWorkspace();
+        }
+
+        void loadModel() throws CoreException {
+            if (buildPathVersioned != null) {
+                return;
+            }
+            loadFileInfo();
+            String contents;
+            try {
+                contents = IO.collect(new InputStreamReader(bndFile.getContents(), bndFile.getCharset()));
+                bndDoc = new Document(contents);
+                bndModel = new BndEditModel(getWorkspace());
+                bndModel.loadFrom(bndDoc);
+            } catch (IOException e) {
+                throw new CoreException(new Status(IStatus.ERROR, BndtoolsConstants.CORE_PLUGIN_ID, "Error trying to read " + bndFile, e));
+            } catch (Exception e) {
+                throw new CoreException(new Status(IStatus.ERROR, BndtoolsConstants.CORE_PLUGIN_ID, "Error trying to access Bnd workspace", e));
+            }
+
+            final List<VersionedClause> origBuildPath = bndModel.getBuildPath();
+            buildPath = origBuildPath == null ? new ArrayList<>() : new ArrayList<>(origBuildPath);
+            final int capacity = 2 * buildPath.size() + 1;
+            buildPathBundles = new HashSet<>(capacity);
+            buildPathVersioned = new HashSet<>(capacity);
+            for (VersionedClause bundleVersion : buildPath) {
+                buildPathBundles.add(bundleVersion.getName());
+                buildPathVersioned.add(bundleVersion);
+            }
+        }
+
+        public List<VersionedClause> getBuildPath() throws CoreException {
+            loadModel();
+            return buildPath;
+        }
+
+        public IFile getBndFile() {
+            loadFileInfo();
+            return bndFile;
+        }
+
+        public void addBundle(String bundle) throws CoreException {
+            addBundle(new VersionedClause(bundle, new Attrs()));
+        }
+
+        public void addBundle(VersionedClause versionedBundle) throws CoreException {
+            loadModel();
+            buildPathBundles.add(versionedBundle.getName());
+            if (buildPathVersioned.add(versionedBundle)) {
+                buildPath.add(versionedBundle);
+            }
+            bndModel.setBuildPath(buildPath);
+            bndModel.saveChangesTo(bndDoc);
+            InputStream str;
+            try {
+                str = new ByteBufferInputStream(bndDoc.get()
+                    .getBytes(bndFile.getCharset()));
+            } catch (UnsupportedEncodingException e) {
+                throw new CoreException(new Status(IStatus.ERROR, BndtoolsConstants.CORE_PLUGIN_ID, "Invalid encoding for " + bndFile, e));
+            }
+            bndFile.setContents(str, IResource.KEEP_HISTORY, monitor);
+        }
+
+        public boolean containsBundle(String bundle) throws CoreException {
+            loadModel();
+            return buildPathBundles.contains(bundle);
+        }
+    }
+
+    class AddBundleJob extends WorkspaceJob {
+        private final String bundle;
+        private final BndBuildPathHandler handler;
+
+        public AddBundleJob(IInvocationContext context, String bundle) {
+            super("Adding '" + bundle + "' to Bnd build path");
+            this.bundle = bundle;
+            handler = getBuildPathHandler(context);
+            IWorkspace workspace = ResourcesPlugin.getWorkspace();
+            IResourceRuleFactory ruleFactory = workspace.getRuleFactory();
+            setRule(ruleFactory.modifyRule(handler.getBndFile()));
+        }
+
+        @Override
+        public IStatus runInWorkspace(IProgressMonitor monitor) {
+            try {
+                handler.setProgressMonitor(monitor);
+                handler.addBundle(bundle);
+
+                return Status.OK_STATUS;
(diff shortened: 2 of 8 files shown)


bndtools.core/src/org/bndtools/core/editors/ImportPackageQuickFixProcessor.java
```diff
@@ -0,0 +1,578 @@
...
+            loadModel();
+            fBuildPathBundles.add(versionedBundle.getName());
+            if (fBuildPathVersioned.add(versionedBundle)) {
+                fBuildPath.add(versionedBundle);
+            }
+            fBndModel.setBuildPath(fBuildPath);
+            fBndModel.saveChangesTo(fBndDoc);
+            InputStream str;
+            try {
+                str = new ByteArrayInputStream(fBndDoc.get()
```
bjhargrave: In general, you want to avoid ByteArray{Input,Output}Stream classes. They are excessively synchronized. We have ByteBuffer{Input,Output}Stream in the aQute.lib.io package as replacements.

bndtools.core/src/org/bndtools/core/editors/ImportPackageQuickFixProcessor.java
```diff
@@ -0,0 +1,578 @@
...
+            }
+            fBndFile.setContents(str, IResource.KEEP_HISTORY, fMonitor);
+        }
+
+        public boolean containsBundle(String bundle) throws CoreException {
+            loadModel();
+            return fBuildPathBundles.contains(bundle);
+        }
+
+        // public boolean containsBundle(VersionedClause versionedBundle) throws CoreException {
```
bjhargrave: Delete the commented out code.

bndtools.core/src/org/bndtools/core/editors/ImportPackageQuickFixProcessor.java
```diff
@@ -0,0 +1,578 @@
...
+        final IInvocationContext fContext;
+
+        public AddBundleCompletionProposal(String bundle, int relevance, IInvocationContext context) {
+            fBundle = bundle;
+            fDisplayString = Strings.format("Add bundle '%s' to Bnd build path", bundle);
+            fRelevance = relevance;
+            fContext = context;
+        }
+
+        // public void printBndFile(String msg, IFile file) {
```
bjhargrave: delete the commented out code

bndtools.core/src/org/bndtools/core/editors/ImportPackageQuickFixProcessor.java
```diff
@@ -0,0 +1,578 @@
...
+        switch (problemId) {
+            case IProblem.ImportNotFound :
+            case IProblem.UndefinedType :
+                return true;
+            default :
+                return false;
+        }
+    }
+
+    protected static Requirement buildRequirement(String importName) {
```
bjhargrave: You use `protected` a lot. Unless this class is designed for subclassing which I do not see that  it is, the just use default or private access for fields and methods.

bndtools.core/src/org/bndtools/core/editors/ImportPackageQuickFixProcessor.java
```diff
@@ -0,0 +1,578 @@
...
+
+        @Override
+        public IStatus runInWorkspace(IProgressMonitor monitor) {
+            try {
+                fHandler.setProgressMonitor(monitor);
+                fHandler.addBundle(fBundle);
+
+                return Status.OK_STATUS;
+            } catch (CoreException e) {
+                return e.getStatus();
```
bjhargrave: I am thinking you may want to log the exception here?

bndtools.core/src/org/bndtools/core/editors/ImportPackageQuickFixProcessor.java
```diff
@@ -0,0 +1,578 @@
...
+
+    List<RepositoryPlugin> listRepositories() {
+        return RepositoryUtils.listRepositories(true);
+    }
+
+    IJavaCompletionProposal[] getSuggestions(Set<String> pkgs, IInvocationContext context) {
+        List<RepositoryPlugin> ps = listRepositories();
+
+        final BndBuildPathHandler handler = getBuildPathHandler(context);
+        SetMultimap<String, String> bundles = MultimapBuilder.treeKeys()
```
bjhargrave: Don't use google collections

bndtools.core/bnd.bnd
```diff
@@ -10,8 +10,10 @@ Bundle-Localization: plugin
 
 # 1. JUNit and Felix Shell are optional dependencies.
 # 2. Disable ALL Eclipse split package attributes, to ensure we import from the "aggregator" bundle(s).
-Import-Package: junit.framework;resolution:=optional,\
+Import-Package: \
+	junit.framework;resolution:=optional,\
 	org.apache.felix.shell;resolution:=optional,\
+	com.google.common.collect;version=16.0.0,\
```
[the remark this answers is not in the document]
bjhargrave: I see that org.bndtools.templating uses google collections because the mustache templating code uses it, but it uses -conditionalpackage to include the used packages as implementation detail to avoid importing the packages causing a runtime dependencies on another bundle.

bndtools.core/src/org/bndtools/core/editors/ImportPackageQuickFixProcessor.java
```diff
@@ -0,0 +1,578 @@
...
+                    continue;
+                }
+                for (Capability cap : caps) {
+                    final String bsn = capabilityToBSN(cap);
+                    try {
+                        if (!handler.containsBundle(bsn)) {
+                            bundles.put(bsn, p.getName());
+                        }
+                    } catch (Exception e) {
+                        e.printStackTrace();
```
bjhargrave: This is not useful in an IDE since no one may see it. Better to log.

bndtools.core/src/org/bndtools/core/editors/ImportPackageQuickFixProcessor.java
```diff
@@ -0,0 +1,578 @@
...
+        return attributes.get(BundleNamespace.BUNDLE_NAMESPACE) + "";
+    }
+
+    /**
+     * Converts a package capability into a VersionedClause.
+     *
+     * @param cap The capability object describing the package being imported.
+     * @return A user-friendly string with the BSN and version.
+     */
+    public static VersionedClause capabilityToVersion(Capability cap) {
```
bjhargrave: This method does not appear to be used.

bndtools.core/src/org/bndtools/core/editors/ImportPackageQuickFixProcessor.java
```diff
@@ -0,0 +1,578 @@
...
+
+            for (Capability providerCap : values) {
+                resultSet.add(providerCap);
+            }
+        }
+        return resultSet;
+    }
+
+    protected static class BndBuildPathHandler {
+        private final IInvocationContext fContext;
```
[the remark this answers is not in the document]
kriegfrj: I copied what appears to be the Eclipse coding standard. Will change to mimic what Bndtools code is doing.

bndtools.core/src/org/bndtools/core/editors/ImportPackageQuickFixProcessor.java
```diff
@@ -0,0 +1,578 @@
...
+import org.eclipse.swt.graphics.Image;
+import org.eclipse.swt.graphics.Point;
+import org.osgi.framework.namespace.BundleNamespace;
+import org.osgi.framework.namespace.PackageNamespace;
+import org.osgi.resource.Capability;
+import org.osgi.resource.Requirement;
+import org.osgi.service.repository.Repository;
+
+import com.google.common.collect.MultimapBuilder;
+import com.google.common.collect.SetMultimap;
```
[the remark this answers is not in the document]
kriegfrj: I was not aware of that - I've changed it to use the Bnd implementation.

One thing to watch will be performance on workspaces with large numbers of repositories, as now the `contains()` call where this was used will now be doing a linear search. With the Guava multimap I had set it up to use a TreeSet as the "bucket" collection type for sorting and more efficient lookup. But the sorting probably isn't that important, and in most workspaces the number of repos will probably be small enough that a linear search will be fine. I didn't notice any significant performance hit in testing. However, if it becomes an issue might need to enhance Bnd's implementation so as to enable specifying alternative collection types for the buckets so that we can have one with Set/fast lookup characteristics.

bndtools.core/src/org/bndtools/core/editors/ImportPackageQuickFixProcessor.java
kriegfrj: Changed to use `ByteBufferInputStream`.

bndtools.core/src/org/bndtools/core/editors/ImportPackageQuickFixProcessor.java
kriegfrj: Done.

bndtools.core/src/org/bndtools/core/editors/ImportPackageQuickFixProcessor.java
kriegfrj: Done.

bndtools.core/src/org/bndtools/core/editors/ImportPackageQuickFixProcessor.java
kriegfrj: Done.

bndtools.core/src/org/bndtools/core/editors/ImportPackageQuickFixProcessor.java
kriegfrj: I had assumed that the Eclipse Job framework would take care of properly reporting the error that is returned via the `IStatus` return value, but given that you asked the question I decided to verify this empirically. I have verified that this is indeed the case - not only will Eclipse log the error, but it will pop up a dialog with information about the error. If we log it manually we'll get duplicates in the error log.

bndtools.core/src/org/bndtools/core/editors/ImportPackageQuickFixProcessor.java
kriegfrj: Fixed.

bndtools.core/bnd.bnd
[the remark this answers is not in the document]
kriegfrj: This is a bit moot now, as I have removed the Guava dependency by changing from the Guava multimap implementation to the Bnd one. However:

1. I would not have pulled in guava were it not for the fact that there was already another bundle in the workspace that was using it.
2. While I can understand dependency fanout being an issue for Bnd, I would argue that it is much less of an issue for Bndtools. The latter is deployed in the context of an Eclipse installation, which already has a large number of dependencies installed (and for this example in particular, Guava is already likely to be among those dependencies, so we wouldn't be adding to the clutter).

bndtools.core/bndtools.shared.bndrun
```diff
@@ -166,6 +166,7 @@ crossplatform.bundles: \
 	${workspace.bundles},\
 	${gogo.bundles},\
 	bndtools.runtime.eclipse.applaunch,\
+	com.google.guava;version=16.0.0,\
```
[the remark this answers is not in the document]
kriegfrj: This has been removed; see above comments about Guava.

bndtools.core/bnd.bnd
```diff
@@ -98,12 +100,16 @@ eclipse.deps: \
 	bndtools.api;version=latest,\
 	org.bndtools.headless.build.manager;version=latest,\
 	org.bndtools.versioncontrol.ignores.manager;version=latest,\
-	org.bndtools.templating; version=latest,\
+	org.bndtools.templating;version=latest,\
 	org.apache.servicemix.bundles.kxml2;version=latest,\
 	org.apache.servicemix.bundles.xmlpull;version=latest,\
-	bndtools.utils;version=project;packages=*
+	bndtools.utils;version=project;packages=*,\
+	com.google.guava;version=16.0
```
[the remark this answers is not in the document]
kriegfrj: This has been removed; see above comments about Guava.

bndtools.core/src/org/bndtools/core/editors/ImportPackageQuickFixProcessor.java
kriegfrj: It was used in an earlier iteration, when I was toying with the idea of showing suggestions for individual available bundle versions, rather than simply suggesting the bare bsn. I left the code in in case I decided to add this functionality again in the future (similarly with `BndBuildPathHandler.containsBundle(VersionedClause)`). I removed it now though - it won't be hard to add again if needed.

bndtools.core/src/org/bndtools/core/editors/ImportPackageQuickFixProcessor.java
kriegfrj: Yes, that was an oversight. I have removed the try/catch and allowed the `CoreException` to propagate up to the Java core through `getCorrections()`. I have verified that Eclipse takes care of reporting the error (to its error log).

bndtools.core/test/org/bndtools/core/editors/FakeFile.java
```diff
@@ -0,0 +1,268 @@
...
+             * Map<Requirement, Collection<Capability>> retval = new HashMap<>(); int i = 0; for (Requirement req :
+             * reqs) { assertThat(req.getNamespace()).as("[" + i + "]namespace")
+             * .isEqualTo(PackageNamespace.PACKAGE_NAMESPACE); assertThat(req.getDirectives()).as("[" + i +
+             * "]directives") .containsOnlyKeys(PackageNamespace.REQUIREMENT_FILTER_DIRECTIVE); final String filter =
+             * req.getDirectives() .get(PackageNamespace.REQUIREMENT_FILTER_DIRECTIVE); assertThat(filter).as("[" + i +
+             * "]filter") .matches("\\(" + PackageNamespace.PACKAGE_NAMESPACE + "=[^)]*\\)"); int start =
+             * filter.indexOf('='); final String pkg = filter.substring(start + 1, filter.length() - 1) .trim(); final
+             * Collection<Capability> caps = thisRepoMap.get(pkg); retval.put(req, (caps == null ?
+             * Collections.emptyList() : caps)); i++; } return retval; } }).when(repo) .findProviders(any());
+             */
```
seanbright: Delete this comment blob

bndtools.core/src/org/bndtools/core/editors/ImportPackageQuickFixProcessor.java
```diff
@@ -0,0 +1,520 @@
...
+            case IProblem.ImportNotFound :
+            case IProblem.UndefinedType :
+                return true;
+            default :
+                return false;
+        }
+    }
+
+    static private Requirement buildRequirement(String importName) {
+        String filter = String.format("(%s=%s)", PackageNamespace.PACKAGE_NAMESPACE, importName.trim());
```
seanbright: I could be wrong, but I feel that we could use one of the `add...()` methods on `CapReqBuilder` to handle this bit. If I'm wrong feel free to ignore.

bndtools.core/bnd.bnd
```diff
@@ -1,4 +1,4 @@
-# Set javac settings from JDT prefs
+	# Set javac settings from JDT prefs
```
seanbright: Revert this whitespace change

bndtools.core/bnd.bnd
```diff
@@ -10,7 +10,8 @@ Bundle-Localization: plugin
 
 # 1. JUNit and Felix Shell are optional dependencies.
 # 2. Disable ALL Eclipse split package attributes, to ensure we import from the "aggregator" bundle(s).
-Import-Package: junit.framework;resolution:=optional,\
+Import-Package: \
```
seanbright: Revert this whitespace change

bndtools.core/bnd.bnd
```diff
@@ -98,12 +99,15 @@ eclipse.deps: \
 	bndtools.api;version=latest,\
 	org.bndtools.headless.build.manager;version=latest,\
 	org.bndtools.versioncontrol.ignores.manager;version=latest,\
-	org.bndtools.templating; version=latest,\
+	org.bndtools.templating;version=latest,\
```
seanbright: Revert this whitespace change

bndtools.core/src/org/bndtools/core/editors/ImportPackageQuickFixProcessor.java
```diff
@@ -0,0 +1,520 @@
...
+import bndtools.central.Central;
+import bndtools.central.RepositoryUtils;
+
+public class ImportPackageQuickFixProcessor implements IQuickFixProcessor {
+
+    private static final ILogger logger = Logger.getLogger(ImportPackageQuickFixProcessor.class);
+
+    // Relevance constants
+    final public static int ADD_BUNDLE = 15;
+    final public static int ADD_BUNDLE_WORKSPACE = ADD_BUNDLE + 1;
```
seanbright: I would prefer the visibility keyword to be first on these.

bjhargrave: accessibility keyword :-)

bjhargrave:
Also, you will need to squash the commits to remove the intermediate states (bug fixes, addressing PR comments).

bndtools.core/src/org/bndtools/core/editors/ImportPackageQuickFixProcessor.java
```diff
@@ -0,0 +1,520 @@
...
+        Set<Capability> resultSet = new LinkedHashSet<>();
+        Map<Requirement, Collection<Capability>> providers = osgiRepo.findProviders(Collections.singleton(req));
+
+        for (Collection<Capability> values : providers.values()) {
+            if (values == null) {
+                continue;
+            }
+
+            for (Capability providerCap : values) {
+                resultSet.add(providerCap);
```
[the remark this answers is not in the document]
kriegfrj: Thanks for highlighting this. I actually managed to simplify it even further by changing `searchRepositories()` to return a `Collection` rather than a `Set`, so that I can return the search results directly. This eliminates the copy completely and is this more efficient.

bndtools.core/test/org/bndtools/core/editors/FakeFile.java
kriegfrj: Done.

bndtools.core/src/org/bndtools/core/editors/ImportPackageQuickFixProcessor.java
kriegfrj: I have managed to change this to use `CapReqBuilder` exclusively, as you suggested.

Note that the implementation of `buildRequirement()` was originally lifted from here: https://github.com/bndtools/bndtools/blob/ca95e3ca7ba2344ec21275eea29b166d1dbc7fcc/bndtools.core/src/bndtools/views/repository/PackageSearchPanel.java#L94-L97

and here:
https://github.com/bndtools/bndtools/blob/ca95e3ca7ba2344ec21275eea29b166d1dbc7fcc/bndtools.core/src/bndtools/views/repository/PackageSearchPanel.java#L105-L110

... so maybe the search code in `bndtools.views.repository` could also be simplified in this way.

bndtools.core/bnd.bnd
kriegfrj: FYI: I don't think I explicitly made these whitespace changes - I think the Bndtools editor did while I was working on it. :smile:

I'm happy to revert, but I'm curious about the reason. Is it because you don't want unrelated changes as part of this commit? Or is it because you liked the whitespace the way it was originally?

bndtools.core/bnd.bnd
kriegfrj: See above.

bndtools.core/bnd.bnd
kriegfrj: See above.

bndtools.core/src/org/bndtools/core/editors/ImportPackageQuickFixProcessor.java
kriegfrj: Fixed. Any chance of this getting added to the code formatter so that it automatically puts all of the modifiers in the correct order (just like it auto-magically formats so much other stuff - whether I want it to or not? 😉 )

kriegfrj:
Ok, I've processed the latest round of suggestions, rebased onto the latest HEAD for master, and squashed all the commits.

I also added another feature - the suggestion will list the repo that the suggested bundle comes from. If the bundle is found in more than one repo, it will name one of the repos and then say " + n other(s)".

It would be nice if you could get Eclipse to show you additional information for quick fixes - as yet, I haven't figured out how to do this, but if it could be done easily, then it could list all of the repos explicitly. Next step after that would be to have options for particular versions.

bndtools.core/bnd.bnd
seanbright: My personal preference is that whitespace or formatting-only changes are not commingled with functional changes. It makes reviewing the functional changes easier.

bjhargrave:
@seanbright Are you ok to merge?

bjhargrave:
Thanks for the PR!

kriegfrj:
No problem!

kriegfrj:
Btw, I think this closes #1408 as well.

PR #1854, opened 2018-03-16
just in case IMessage[] has null messages

gamerson:


bndtools.core/src/bndtools/utils/MessageHyperlinkAdapter.java
```diff
@@ -47,12 +47,22 @@ private void showPopup(final HyperlinkEvent e) {
 
         IMessage[] messages = (IMessage[]) e.data;
 
-        if (messages == null || messages.length == 0) {
+        if (messages == null || messages.length == 0 || hasNull(messages)) {
             MessageDialog.openInformation(part.getSite()
```
bjhargrave: Wouldn't it be better to filter out the nulls (once you have confirmed messages != null)?

```java
messages = Stream.of(messages).filter(Objects::nonNull).toArray(IMessage[]::new);
```

Then the non-null messages are displayed.

gamerson: agreed

PR #1855, opened 2018-03-16
add to run instead of workspace

gamerson:


bndtools.m2e/src/bndtools/m2e/MavenWorkspaceRunListener.java
```diff
@@ -10,14 +10,16 @@
 
     @Override
     public void create(Run run) throws Exception {
-        run.getWorkspace()
-            .addBasicPlugin(new MavenWorkspaceRepository());
+        MavenWorkspaceRepository repo = run.getPlugin(MavenWorkspaceRepository.class);
```
bjhargrave: I am not sure this will work. I still think you need to add the plugin to the Run's workspace. But given the other changes above, the workspace should be a standalone workspace.

gamerson: updated.

PR #1872, opened 2018-03-27
provide osgi.service capability

gamerson:


bndtools.builder/src/org/bndtools/builder/decorator/ui/ProjectDecoratorImpl.java
```diff
@@ -3,7 +3,12 @@
 import org.bndtools.build.api.IProjectDecorator;
 import org.bndtools.builder.ComponentMarker;
 import org.eclipse.core.resources.IProject;
+import org.osgi.annotation.bundle.Capability;
+import org.osgi.namespace.service.ServiceNamespace;
 
+@Capability(namespace = ServiceNamespace.SERVICE_NAMESPACE, attribute = {
+    "objectClass:List<String>=\"org.bndtools.build.api.IProjectDecorator\""
+})
```
bjhargrave: You should add `uses=org.bndtools.build.api.IProjectDecorator.class`.

gamerson: Updated but this causes a weird error in Bndtools.

PR #1878, opened 2018-04-24
[m2e] always put the maven repo first in the list

rotty3000:


bndtools.m2e/src/bndtools/m2e/MavenWorkspaceRunListener.java
```diff
@@ -15,7 +15,10 @@ public void create(Run run) throws Exception {
         MavenWorkspaceRepository repo = workspace.getPlugin(MavenWorkspaceRepository.class);
 
         if (repo == null) {
-            workspace.addBasicPlugin(new MavenWorkspaceRepository());
+            MavenWorkspaceRepository mavenWorkspaceRepository = new MavenWorkspaceRepository();
+            workspace.addBasicPlugin(mavenWorkspaceRepository);
+            workspace.getRepositories()
+                .add(0, mavenWorkspaceRepository);
```
bjhargrave: Now the repository is in the list twice. Perhaps you should remove it from the list before you add it at the head of the list.

rotty3000:
See https://github.com/bndtools/bndtools/pull/1879

bndtools.m2e/src/bndtools/m2e/MavenWorkspaceRunListener.java
rotty3000: I believe the `getRepositories` has an internal cache that once created is never overridden by a call to `addBasicPlugin` and I actually meant to make that call second, but I got overexcited.

PR #1899, opened 2018-06-19
Removing dead code. Fixes #1898

### LINKED ISSUE #1898: Remove dead code
I'm pretty sure that the code in this package hasn't been used for a long time (since the templating engine was moved to use bundles in the workspace). I can't find any references to it in the rest of bndtools.

https://github.com/bndtools/bndtools/tree/master/bndtools.core/src/org/bndtools/core/templates/enroute

timothyjward:
Signed-off-by: Tim Ward <timothyjward@apache.org>

Adding @bjhargrave and @pkriens to verify that this code really is unused.


pkriens:
Yes, it can go. This was already replaced by Neil's work

PR #1919, opened 2018-08-13
[m2e] Refresh only the target JAR, not the whole target dir

njbartlett:
Patch contributed by Simon Chemouil <schemouil@gmail.com>. In
BndConfigurator we were refreshing the whole Maven `target` dir after
calling the `jar:jar` goal. In some cases this triggered a rebuild of
the project, resulting in an infinite loop. By refreshing only the
target JAR file, this seems to be fixed.

Signed-off-by: Neil Bartlett <njbartlett@gmail.com>

### DIFF
--- a/bndtools.m2e/src/bndtools/m2e/BndConfigurator.java
+++ b/bndtools.m2e/src/bndtools/m2e/BndConfigurator.java
@@ -9,12 +9,15 @@ import java.util.Set;
 import java.util.concurrent.atomic.AtomicReference;
 
 import org.apache.maven.lifecycle.MavenExecutionPlan;
+import org.apache.maven.model.Plugin;
 import org.apache.maven.plugin.MojoExecution;
 import org.apache.maven.project.MavenProject;
 import org.bndtools.api.ILogger;
 import org.bndtools.api.Logger;
 import org.bndtools.build.api.IProjectDecorator;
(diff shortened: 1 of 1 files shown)


magnet:
Hey,
Thanks @njbartlett for taking the patch in.

I took the target file logic from https://github.com/bndtools/bndtools/blob/master/bndtools.m2e/src/bndtools/m2e/MavenWorkspaceRepository.java#L91

I would suggest making these methods (returning a String of the filename or the file itself) static in a helper class to avoid spreading the logic. I copied it because it was easier without an IDE to try the fix that way.

@bjhargrave it should mean that the issue you report already exists I suppose.

ViToni:
Wondering if something similar might happen also with regular bndtools since sometimes I have the impression that a lot of refreshing is happening on a full build.

njbartlett:
I agree with BJ. I opened this PR on Simon's behalf since I believe he was travelling and didn't have full access to GitHub.

It would be better if he opened the PR himself with "Signed-off By". @magnet Could you do that please? Thanks.

timothyjward:
@magnet please make sure that this doesn't interfere with the Automatic regeneration of indexes from the bnd-indexer-maven plugin, which also need to be triggered in dependent projects whenever a jar is rebuilt.

PR #1934, opened 2018-09-12
build: Fix warning from Bnd about fragment host

bjhargrave:
`warning: Host biz.aQute.bndlib=bundle-version="[4.1.0,5.0.0)" for this fragment cannot be found on the classpath`


pkriens:
Wow, it worked! :-)

PR #1947, opened 2018-11-01
fixes #1922 make the test classpath not transitive

### LINKED ISSUE #1922: Runtime Classpath is transitive
I ran into a problem in bnd that the JUnit runner is calculating the class path transitively. Looking at the code, BJ found BndContainerRuntimeClasspathEntryResolver. This class was added two years ago by @gamerson and is clearly transitive.

Having a transitive classpath for JUnit is convenient (as always) but it also means you get a lot of unwanted dependencies since it looks like we do not distinguish between test and build paths?

So do we really want transitive dependencies? The 'native' launcher from bnd does not have transitive dependencies.

rotty3000:


rotty3000:
@pkriens does this solve it for you?

rotty3000:
guys shall I merge this? @bjhargrave @pkriens

pkriens:
Let's try it ...

pkriens:
this removes the whole function, shouldn't you make it conditional on m2e projects?

rotty3000:
@pkriens I don't have time today to deeply check, so either I remove it now and check back later or we leave it and you continue to suffer. Your choice. The simplest fix if m2e really needs it is revert deleting the class and putting the plugin in the `bndtools.m2e` extension instead of `bndtools.core` which will only cause it to be applied to m2e projects... but again, I want to make sure that even makes sense.

pkriens:
I am not sure I understand what is going on. How does this then work in plain JUnit? Thinking about I think we can just remove it? If it has a function in m2e then m2e would fail, which it likely does not?

I am not in a terrible hurry but it creates bad practices ... People rely on transitive deps because it is so easy but then create another tangled web.

pkriens:
Thanks for the effort btw.

PR #1949, opened 2018-11-01
Fixes #1948 updated to latest m2e framework results in endlessly building project

### LINKED ISSUE #1948: updated to latest m3e framework results in endlessly building project
The latest m2e is slightly more pedantic about propagating warnings.

You may end up in a scenario where you have a build _warning_, which `bndtools.m2e` ignores and allows the `jar:jar` task to be executed, finalized by a project refresh, which results in the build firing the _warning_, and so on in an endless loop.

rotty3000:


gamerson:
hey @rotty3000  can you show me a reproducible test case and maybe I could have a look? This fix seems it may be papering over something important.

